PR #1, opened 2016-07-06
blobutil: parse blob info from url

ahmetalpbalkan:
cc: @boumenot

### DIFF
--- a/blobutil/url.go
+++ b/blobutil/url.go
@@ -1,0 +1,76 @@
+package blobutil
+
+import (
+	"fmt"
+	"net/url"
+	"strings"
+
+	"github.com/pkg/errors"
+)
+
+// AzureBlobRef contains information parsed from an Azure Blob Storage URL.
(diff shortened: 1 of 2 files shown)


blobutil/url.go
```diff
@@ -0,0 +1,76 @@
...
+	Blob string
+
+	// Scheme contains http or https
+	Scheme string
+}
+
+// ParseBlobURL recognizes a given Azure Blob Storage URL and extracts the information
+// of the blob such as storage API endpoint, container and blob name, or returns error if the
+// URL is unrecognized.
+func ParseBlobURL(blobURL string) (v AzureBlobRef, err error) {
```
boumenot: Is the SDK every going to handle this stuff for you?

boumenot:
LGTM.

blobutil/url.go
ahmetalpbalkan: @boumenot sdk does not handle this.

PR #2, opened 2016-07-06
preprocess: utils to deal with BOM, DOS, UTF-16

ahmetalpbalkan:
Contains the utilities to detect & sanitize text files (i.e. scripts).

- bom.go: removal of bom chars, utf16 to utf8 conversion
- dos.go: removal of DOS line endings
- file.go: detection of script files via file ext or shebang

and bunch of test files (inherited from old repo)


preprocess/bom.go
```diff
@@ -0,0 +1,49 @@
...
+)
+
+var bomSequences = [][]byte{
+	{'\xef', '\xbb', '\xbf'}, // python: codecs.BOM_UTF8
+	{'\xff', '\xfe'},         // python: codecs.BOM, codecs.BOM_LE, codecs.BOM_UTF16_LE
+	{'\xfe', '\xff'},         // python: codecs.BOM_BE, codecs.BOM_UTF16_BE
+}
+
+var (
+	utf16le = []byte{'\xff', '\xfe'}
```
boumenot: utf8 too?

ahmetalpbalkan: ah I'm realizing these are leftover..

preprocess/file.go
```diff
@@ -0,0 +1,63 @@
...
+	"unicode"
+
+	"github.com/pkg/errors"
+)
+
+const peekLen = 64 // look at first N bytes to figure out if it has shebang
+
+// textExtensions is predefined list of script and text
+// file extensions.
+var textExtensions = []string{
```
boumenot: is this from the current version?  .rb for ruby?

preprocess/file_test.go
```diff
@@ -0,0 +1,47 @@
...
+	require.True(t, hasShebang([]byte("#! foo")), "shebang")
+
+	script := "\r\n  \t#!foo"
+	require.True(t, hasShebang([]byte(script)), "whitespace")
+
+	noscript := "    !$"
+	require.False(t, hasShebang([]byte(noscript)), "not shebang")
+
+	bomdata := append(bomSequences[1], []byte("    \t \n#!booyah")...)
+	require.True(t, hasShebang(bomdata), true)
```
boumenot: I would have said false.  Shebang should be at the absolute beginning, minus the BOM.  (Did I interpret that correctly?)

boumenot:
Few questions, but LGTM.

preprocess/file.go
ahmetalpbalkan: they are mostly from the current version. I think I just added .txt and .pl. Can be extended but I'd say let's not if there's no clear need.

PR #3, opened 2016-07-06
download: utils to download files and blobs

ahmetalpbalkan:
This one is fairly simple:

```

package download

FUNCTIONS

func Download(d Downloader) (io.ReadCloser, error)
    Download retrieves a response body and checks the response status code
    to see if it is 200 OK and then returns the response body. It issues a
    new request every time called. It is caller's responsibility to close
    the response body.

TYPES

type Downloader interface {
    // GetRequest returns a new GET request for the resource.
    GetRequest() (*http.Request, error)
}
    Downloader describes a method to download files.

func NewBlobDownload(accountName, accountKey string, blob blobutil.AzureBlobRef) Downloader
    NewBlobDownload creates a new Downloader for a blob hosted in Azure Blob
    Storage.

func NewURLDownload(url string) Downloader
    NewURLDownload creates a new downloader with the provided URL

```

### DIFF
--- a/download/blob_test.go
+++ b/download/blob_test.go
@@ -1,0 +1,130 @@
+package download
+
+import (
+	"fmt"
+	"io/ioutil"
+	"math/rand"
+	"os"
+	"testing"
+	"time"
+
+	"github.com/Azure/azure-sdk-for-go/storage"
+	"github.com/ahmetalpbalkan/custom-script-extension-linux/blobutil"
+	"github.com/stretchr/testify/require"
+)
+
+func Test_blobDownload_validateInputs(t *testing.T) {
+	type sas interface {
+		getURL() (string, error)
+	}
+
+	_, err := NewBlobDownload("", "", blobutil.AzureBlobRef{}).GetRequest()
+	require.NotNil(t, err)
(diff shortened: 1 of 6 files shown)


download/blob_test.go
```diff
@@ -0,0 +1,130 @@
...
+import (
+	"fmt"
+	"io/ioutil"
+	"math/rand"
+	"os"
+	"testing"
+	"time"
+
+	"github.com/Azure/azure-sdk-for-go/storage"
+	"github.com/ahmetalpbalkan/custom-script-extension-linux/blobutil"
```
boumenot: ahmetalpbalkan -> Azure?

boumenot:
LGTM

download/blob_test.go
ahmetalpbalkan: I'll update these altogether.

ahmetalpbalkan:
@boumenot actually one thing to note is, in this rewrite I switched from authenticating to blob storage and getting the blob (via the Authorization header crafted by the storage client) to downloading the blob from a SAS URL (crafted by the storage client) that is generated freshly every time and is good 30 minutes. I do not foresee any issues with this, let me know if you can think of any. They should work the same.

PR #4, opened 2016-07-06
json-schema checking for handler settings

ahmetalpbalkan:
These utility functions use json-schema (https://tools.ietf.org/html/draft-zyp-json-schema-04)
to validate and if necessary, error out when users specify extra
fields, wrong types, omit required fields etc.

This will save us many validation checks in the code, such as if the required fields are there,
if the inputs match the required format and so on.

### DIFF
--- a/schema.go
+++ b/schema.go
@@ -1,0 +1,102 @@
+package main
+
+import (
+	"fmt"
+
+	"github.com/pkg/errors"
+	"github.com/xeipuuv/gojsonschema"
+)
+
+// Refer to http://json-schema.org/ on how to use JSON Schemas.
+
+const (
+	publicSettingsSchema = `{
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "title": "Custom Script - Public Settings",
+    "type": "object",
+    "properties": {
+        "commandToExecute": {
+            "description": "Command to be executed",
+            "type": "string"
+        },
+        "fileUris": {
+            "description": "List of files to be downloaded",
+            "type": "array",
+            "items": {
+                "type": "string",
+				"format": "uri"
+            }
+        },
+		"timestamp": {
+			"description": "An integer, intended to trigger re-execution of the script when changed",
+			"type":"integer" 
+		}
+    },
+	"additionalProperties": false,
+    "required": ["commandToExecute"]
+}`
+
+	protectedSettingsSchema = `{
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "title": "Custom Script - Protected Settings",
+    "type": "object",
+    "properties": {
+        "commandToExecute": {
+            "description": "Command to be executed",
+            "type": "string"
+        },
+        "storageAccountName": {
+            "description": "Name of the Azure Storage Account (3-24 characters of lowercase letters or digits)",
+            "type": "string",
+			"pattern": "^[a-z0-9]{3,24}$"
+        },
+		"storageAccountKey": {
+            "description": "Key for the Azure Storage Account (a base64 encoded string)",
+            "type": "string",
+			"pattern": "^(?:[A-Za-z0-9+/]{4})*(?:[A-Za-z0-9+/]{2}==|[A-Za-z0-9+/]{3}=|[A-Za-z0-9+/]{4})$"
+        }
+    },
+	"additionalProperties": false
+}`
+)
+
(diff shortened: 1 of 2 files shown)


schema.go
```diff
@@ -0,0 +1,102 @@
...
+	"fmt"
+
+	"github.com/pkg/errors"
+	"github.com/xeipuuv/gojsonschema"
+)
+
+// Refer to http://json-schema.org/ on how to use JSON Schemas.
+
+const (
+	publicSettingsSchema = `{
```
boumenot: Nice!

schema.go
```diff
@@ -0,0 +1,102 @@
...
+    },
+	"additionalProperties": false
+}`
+)
+
+// validateObjectJSON validates the specified json with schemaJSON.
+// If json is empty string, it will be converted into an empty JSON object
+// before being validated.
+func validateObjectJSON(schema *gojsonschema.Schema, json string) error {
+	if json == "" {
```
boumenot: Why?  Seems odd to set json to "{}", why don't you just return ASAP.

schema_test.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+
+	// empty
+	chkPatternMismatch(validateProtectedSettings(`{"storageAccountKey": ""}`), "empty")
+
+	// bad string
+	chkPatternMismatch(validateProtectedSettings(`{"storageAccountKey": "NotABase64Really!"}`), "not b64")
+
+	// missing '=' at the end
+	chkPatternMismatch(validateProtectedSettings(`{"storageAccountKey": "OllwYfXmC0mSMhWg4x+lUdLg6Eoa/d44+PxPTXBaadO5l87L4JzgkyyVvQr8r60WIzG2X8r6LLxkhNBQaHa3XQ="}`), "bad b64")
```
boumenot: Is it always true that base64 has two equals at the end?  I did not think so.

schema_test.go
```diff
@@ -0,0 +1,121 @@
...
+	require.Nil(t, validateProtectedSettings("{}"), "empty string")
+}
+
+func TestValidateProtectedSettings_unrecognizedField(t *testing.T) {
+	err := validateProtectedSettings(`{"alien":0}`)
+	require.NotNil(t, err)
+	require.Contains(t, err.Error(), "Additional property alien is not allowed")
+}
+
+func TestValidateProtectedSettings_commandToExecute(t *testing.T) {
```
boumenot: Too bad these validations aren't at the time of submission.

boumenot:
LGTM.

schema.go
ahmetalpbalkan: @boumenot because it does not necessarily make it valid. consider empty public settings json. It does not contain `commandToExecute`; so it's not valid and it has to be validated.

However if I pass the empty string to `schema.Validate` it'll error out saying, `EOF` because empty string is not valid JSON at all (`""` is valid json, `null` is valid json, zero bytes isn't).

Therefore instead of passing empty string (which would give me an error like: `expected: object, got: string`, because public settings should be a JSON object), now I convert empty string into `{}` and pass it to validation and it tells me `required field empty: commandToExecute`.

schema_test.go
ahmetalpbalkan: you mean at the Set Extension API?

schema_test.go
ahmetalpbalkan: @boumenot you're right this is not a very good test. azure storage account keys are always the same length, so the last group in base64 will always contain 16 bits, therefore removing the last `=`  in this string makes it invalid.

However the length of the account key is not covered by any guarantee, so I will remove this particular line.

ahmetalpbalkan: or maybe we should keep this test because it actually validates that given a string of appropriate length, removing the `=` is caught by this test. If somebody changes the regex in the schema to a very simple regex that appears to be covering base64, this test will catch it!

schema.go
```diff
@@ -0,0 +1,102 @@
...
+		"timestamp": {
+			"description": "An integer, intended to trigger re-execution of the script when changed",
+			"type":"integer" 
+		}
+    },
+	"additionalProperties": false,
+    "required": ["commandToExecute"]
+}`
+
+	protectedSettingsSchema = `{
```
ahmetalpbalkan: lots of tabs vs spaces mix in these embedded schemas, I will fix.

PR #9, opened 2016-07-07
parseCmd: shlex functionality

ahmetalpbalkan:
There's this https://github.com/flynn-archive/go-shlex/ package that is a fork of Google's go-shlex that got deprecated. (Apache 2.0 LICENSE).

Added a wrapper function for that and tests around it. Also vending the source code for that.

### DIFF
--- a/cmd.go
+++ b/cmd.go
@@ -1,7 +1,7 @@
 package main
 
 import (
-	"github.com/flynn-archive/go-shlex"
+	"github.com/google/shlex"
 	"github.com/pkg/errors"
 )
 
--- a/cmd.go
+++ b/cmd.go
@@ -1,0 +1,14 @@
+package main
+
+import (
+	"github.com/flynn-archive/go-shlex"
+	"github.com/pkg/errors"
+)
+
+// parseCmd is a wrapper function for the underlying shell lexer implementation,
+// it splits a commands string into argv slice based upon shell quoting and escaping
+// rules.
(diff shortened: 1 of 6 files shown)


cmd.go
```diff
@@ -0,0 +1,14 @@
...
+	"github.com/flynn-archive/go-shlex"
+	"github.com/pkg/errors"
+)
+
+// parseCmd is a wrapper function for the underlying shell lexer implementation,
+// it splits a commands string into argv slice based upon shell quoting and escaping
+// rules.
+func parseCmd(cmd string) ([]string, error) {
+	v, err := shlex.Split(cmd)
+	return v, errors.Wrap(err, "shlex: failed to parse command")
```
boumenot: What's the value is parsing the shell command, as opposed to just passing it through?  Are you trying to find more errors, and fail faster?

boumenot:
LGTM.

cmd.go
ahmetalpbalkan: @boumenot the way you pass a command directly to OS can be through `/bin/bash -c $CMD` and in that case you are giving shell access to the program. It is basically a security flaw. Also doing shlex is preserving the existing behavior.

ahmetalpbalkan:
I'm switching the vendored package to `github.com/google/shlex` directly. apparently this fork is not maintained and the guy just closed the issues section on the repo to my face. screw the Go package ecosystem. 
image

PR #10, opened 2016-07-07
Revert cmd/shlex implementation

ahmetalpbalkan:
Per offline discussion, decided to not to do shlex but /bin/sh -c
instead in favor of a better lexer and lesser dependencies.


boumenot:
LGTM

PR #13, opened 2016-07-12
exec: helper method to run cmds and save outputs

ahmetalpbalkan:
ExecCmdInDir executes the given command in given directory and saves output to ./stdout and ./stderr files (truncates files if exists, creates them if not with 0600/-rw------- permissions).

Ideally, we execute commands only once per sequence number in custom-script-extension, and save their output under `/var/lib/waagent/<dir>/download/<seqnum>/*`.

### DIFF
--- a/exec.go
+++ b/exec.go
@@ -3,7 +3,9 @@ package main
 import (
 	"fmt"
 	"io"
+	"os"
 	"os/exec"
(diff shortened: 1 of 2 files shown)


boumenot:
LGTM.

PR #14, opened 2016-07-13
download: add WithRetries method

ahmetalpbalkan:
WithRetries will retry any download failure and will sleep
in exponentially increasing durations up to 5 min. (5s, 10s,
20s, 40s, 80s, 160s)

The existing extension sleeps a fixed amount and up to 10 minutes.
It also retries all errors (bad host, 404s etc). I kept that part
the same, can be improved with an IsRetryable func later on perhaps.

### DIFF
--- a/download/retry_test.go
+++ b/download/retry_test.go
@@ -1,0 +1,101 @@
+package download_test
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/Azure/custom-script-extension-linux/download"
+	"github.com/ahmetalpbalkan/go-httpbin"
+	"github.com/stretchr/testify/require"
+)
+
+var (
+	// how much we sleep between retries
+	sleepSchedule = []time.Duration{
+		5 * time.Second,
+		10 * time.Second,
+		20 * time.Second,
+		40 * time.Second,
+		80 * time.Second,
+		160 * time.Second}
+)
+
+func TestActualSleep_actuallySleeps(t *testing.T) {
+	s := time.Now()
+	download.ActualSleep(time.Second)
+	e := time.Since(s)
+	require.InEpsilon(t, 1.0, e.Seconds(), 0.01, "took=%fs", e.Seconds())
+}
+
+func TestWithRetries_noRetries(t *testing.T) {
+	srv := httptest.NewServer(httpbin.GetMux())
+	defer srv.Close()
+
+	d := download.NewURLDownload(srv.URL + "/status/200")
+
+	sr := new(sleepRecorder)
+	resp, err := download.WithRetries(d, sr.Sleep)
+	require.Nil(t, err, "should not fail")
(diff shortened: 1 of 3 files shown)


download/retry_test.go
```diff
@@ -0,0 +1,101 @@
...
+		5 * time.Second,
+		10 * time.Second,
+		20 * time.Second,
+		40 * time.Second,
+		80 * time.Second,
+		160 * time.Second}
+)
+
+func TestActualSleep_actuallySleeps(t *testing.T) {
+	s := time.Now()
```
boumenot: I'm not sure how important this really is.  If you're spending < 1s executing this test case I don't care that much.

ahmetalpbalkan: this is because somebody perhaps changes `var ActualSleep` with some mock which doesn't actually sleep like wall clock. it'll detect that side-effect.

download/retry_test.go
```diff
@@ -0,0 +1,101 @@
...
+
+func TestWithRetries_failingBadStatusCode_validateSleeps(t *testing.T) {
+	srv := httptest.NewServer(httpbin.GetMux())
+	defer srv.Close()
+
+	d := download.NewURLDownload(srv.URL + "/status/404")
+
+	sr := new(sleepRecorder)
+	_, err := download.WithRetries(d, sr.Sleep)
+	require.EqualError(t, err, "unexpected status code: got=404 expected=200")
```
boumenot: You're downloading from /status/404, but you expected 200.  It's a little confusing.

I assume /status/404 always returns 404?

download/retry_test.go
```diff
@@ -0,0 +1,101 @@
...
+// sleepRecorder keeps track of the durations of Sleep calls
+type sleepRecorder []time.Duration
+
+// Sleep does not actually sleep. It records the duration and returns.
+func (s *sleepRecorder) Sleep(d time.Duration) {
+	*s = append(*s, d)
+}
+
+// healingServer returns HTTP 500 until 4th call, then HTTP 200 afterwards
+type healingServer int
```
boumenot: Nice name.

boumenot:
Just one question, but otherwise LGTM.

download/retry_test.go
ahmetalpbalkan: @boumenot so `Download()` function expects 200 OK responses but this URL will always 404 and hence the retries will always fail and I am expecting an error like `unexpected status code: got=404 expected=200`

ahmetalpbalkan:
just clarified, lmk if sounds good.

download/retry_test.go
boumenot: Makes more sense.  Is there any need to revalidate the error message in this case?  I think it sufficient to assert you retried.

boumenot:
LGTM.

download/retry_test.go
ahmetalpbalkan: well, just to make sure if I'm returning the original error back.

PR #15, opened 2016-07-18
download/save: save downloaded urls as files

ahmetalpbalkan:


download/save.go
```diff
@@ -0,0 +1,30 @@
+package download
+
+import "os"
+import "github.com/pkg/errors"
+import "io"
```
boumenot: Fix order.

download/save.go
```diff
@@ -0,0 +1,30 @@
...
+
+const (
+	writeBufSize = 1024 * 2014
```
boumenot: I would have thought 8k or 16k, but that's an odd choice.

download/save_test.go
```diff
@@ -0,0 +1,84 @@
...
+	require.Nil(t, err)
+
+	fi, err := os.Stat(path)
+	require.Nil(t, err)
+	require.EqualValues(t, 128, fi.Size())
+	require.Equal(t, os.FileMode(0600), fi.Mode().Perm(), "mode should not be changed")
+}
+
+func TestSave_largeFile(t *testing.T) {
+	srv := httptest.NewServer(httpbin.GetMux())
```
boumenot: How long does it take to run this test?

boumenot:
Few questions, but LGTM.

download/save.go
ahmetalpbalkan: Hmm, I could use advice. I was planning to use io.CopyBuffer and let go decide the buffer size. It appears like the old code has 1 KB of buffer size.

download/save_test.go
ahmetalpbalkan: Roughly 2.7-3s regardless of buffer size being 1k or 1mb. When I save to /dev/null, it's about 0.5s. So I think it is  pretty much the disk write speed on my laptop?

download/save.go
boumenot: My vote is 8/16k.

ahmetalpbalkan: done.

download/save_test.go
ahmetalpbalkan: travis build is faster:

https://travis-ci.org/Azure/custom-script-extension-linux/builds/145643847

=== RUN   TestSave_largeFile
--- PASS: TestSave_largeFile (0.39s)

PR #16, opened 2016-07-19
Process file urls

ahmetalpbalkan:
These methods will help download a file from URL,
save it to a directory with proper name and apply
post-download/pre-execution processing (DOS, BOM etc)
on it.

### DIFF
--- a/files.go
+++ b/files.go
@@ -22,10 +22,7 @@ func downloadAndProcessURL(url, downloadDir, storageAccountName, storageAccountK
 		return err
 	}
 
-	dl, err := getDownloader(
-		url,
-		storageAccountName,
-		storageAccountKey)
+	dl, err := getDownloader(url, storageAccountName, storageAccountKey)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


files.go
```diff
@@ -0,0 +1,99 @@
...
+		return err
+	}
+
+	fp := filepath.Join(downloadDir, fn)
+	const mode = 0500 // we assume users download scripts to execute
+	if _, err := download.SaveTo(dl, fp, mode); err != nil {
+		return err
+	}
+
+	err = postProcessFile(fp)
```
boumenot: Why don't you pass in a func() instead of hard coding postProcessFile(). It _seems_ like a good extension point, but maybe it isn't necessary yet.

files.go
```diff
@@ -0,0 +1,99 @@
...
+	return errors.Wrapf(err, "failed to post-process '%s'", fn)
+}
+
+// getDownloader returns a downloader for the given URL based on whether the
+// storage credentials are empty or not.
+func getDownloader(fileURL string, storageAccountName, storageAccountKey string) (
+	download.Downloader, error) {
+	if storageAccountName != "" && storageAccountKey != "" {
```
boumenot: You can save 10 lines of indent if you invert your if statement.

files.go
ahmetalpbalkan: not sure if I understood that fully.

files.go
ahmetalpbalkan: makes sense

files.go
```diff
@@ -0,0 +1,99 @@
...
+			storageAccountKey,
+			blob), nil
+	}
+	return download.NewURLDownload(fileURL), nil
+}
+
+// urlToFileName parses given URL and returns the section after the last slash
+// character of the path segment to be used as a file name. If a value is not
+// found, an error is returned.
+func urlToFileName(fileURL string) (string, error) {
```
boumenot: Why don't you start with url.URL in the first place?  Is this the first time the fileURL is handled?

files_test.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	for _, c := range cases {
+		_, err := urlToFileName(c)
+		require.NotNil(t, err, "not failed: %s", "url=%s", c)
+		require.Contains(t, err.Error(), "cannot extract file name from URL", "url=%s", c)
+	}
+}
+
+func Test_urlToFileName(t *testing.T) {
```
boumenot: Nice.

boumenot:
Few questions, but LGTM.

files.go
ahmetalpbalkan: pretty much. we don't need the *url.URL anywhere else at all. Also `url.Parse` does not mean that the passed url is a full URL, it can be empty string, a relative url (e.g. `foo`) or a full url. We don't know for sure if the url is full (i.e. fetchable) until we attempt to download the file via http.Client.Get().

files.go
boumenot: Change the method to accept a lambda.  Something like this.

```go
func downloadAndProcessURL([snip], preprocess func(filename string) error) error { 
```

ahmetalpbalkan: hmm probably not so necessary. not all preprocess funcs operate on filename.

ahmetalpbalkan:
addressed CR comments, will merge..

PR #18, opened 2016-07-20
seqnum: sequence number tracking via file

ahmetalpbalkan:
I gave a thought about using/extending docker-extension's
seqnumfile pkg but it has some hardcoded stuff and designed to
prevent executing the sequence number simultaneously.

This code is shorter and exposes only methods to make sure
we only process sequence numbers greater than the stored one.

### DIFF
--- a/seqnum/seqnum.go
+++ b/seqnum/seqnum.go
@@ -1,0 +1,39 @@
+// Package seqnum provides utilities to persistently store sequence number in a
+// file and compare against the stored value to ensure a sequence number is
+// processed only once in the extension handler.
+package seqnum
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"strconv"
+
+	"github.com/pkg/errors"
+)
+
+const (
+	// chmod is used to set the mode bits for new seqnum files.
+	chmod = os.FileMode(0600)
+)
+
+// Set replaces the stored sequence number in file, or creates a new file at
+// path if it does not exist.
+func Set(path string, num int) error {
+	b := []byte(fmt.Sprintf("%v", num))
+	return errors.Wrap(ioutil.WriteFile(path, b, chmod), "seqnum: failed to write")
+}
+
+// IsSmallerThan returns if the sequence number stored at path is smaller than
+// the provided num. If no number is stored, returns true and no error.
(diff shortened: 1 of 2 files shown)


seqnum/seqnum.go
```diff
@@ -0,0 +1,39 @@
...
+func IsSmallerThan(path string, num int) (bool, error) {
+	b, err := ioutil.ReadFile(path)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return true, nil
+		}
+		return false, errors.Wrap(err, "seqnum: failed to read")
+	}
+	stored, err := strconv.Atoi(string(b))
+	return stored < num, errors.Wrapf(err, "seqnum: cannot parse number %q", b)
```
boumenot: I assume errors.Wrapf(...) returns nil if err is nil?

seqnum/seqnum_test.go
```diff
@@ -0,0 +1,113 @@
...
+
+	require.Nil(t, seqnum.Set(fp, 1))
+
+	b, err := seqnum.IsSmallerThan(fp, 2)
+	require.Nil(t, err)
+	require.True(t, b, "1<2")
+
+	b, err = seqnum.IsSmallerThan(fp, 1)
+	require.Nil(t, err)
+	require.False(t, b, "1≮1")
```
boumenot: Nice symbol (≮).  What is it?

boumenot:
LGTM.

Can seqno be folded into the global library for extensions at some point?

seqnum/seqnum.go
ahmetalpbalkan: yes

ahmetalpbalkan:
> Can seqno be folded into the global library for extensions at some point?

@boumenot maybe. I'm not sure if we have other extensions that need to run only once ever. If they need it, they can use it from here just as I use some stuff from docker-extension. If the need for writing extensions in Go grows bigger, we can extract into a separate repo.

seqnum/seqnum_test.go
ahmetalpbalkan: @boumenot your font might be tricking you.

image

boumenot:
SGTM.  This was kind of my feeling too.

PR #19, opened 2016-07-20
download/retry: reduce retry duration

ahmetalpbalkan:
the old agent retries constant time for 10*20s=200s. modifying
this to retry for total 189s. to be closer to 3 minutes instead
of minutes.


boumenot:
LGTM.

PR #21, opened 2016-07-20
preprocess: increase code coverage

ahmetalpbalkan:


preprocess/bom_test.go
```diff
@@ -58,3 +58,17 @@ func TestRemoveBOM(t *testing.T) {
 		require.True(t, bytes.HasPrefix(n, []byte("#!")), "%s does not start with shebang: %#v", fn, n)
 	}
 }
+
+func Test_encodeToUTF8(t *testing.T) {
```
boumenot: This feels like a rat hole. 😦 So many ways to be wrong, and we are very forgiving.

boumenot:
LGTM.

preprocess/bom_test.go
ahmetalpbalkan: @boumenot in case I'd say "best effort" instead of forgiving. if we cannot determine the encoding to be utf8 or utf16, we just preserve the original encoding.

PR #22, opened 2016-07-20
files_test.go: increase coverage

ahmetalpbalkan:
it's a bit hard to test error cases in here (as it'll trigger 2 minute retries etc)


boumenot:
LGTM.

PR #23, opened 2016-07-20
main.go: starting to write main.go

ahmetalpbalkan:
Some skeleton code for starting to write main executable.

### DIFF
--- a/cmds.go
+++ b/cmds.go
@@ -6,9 +6,9 @@ type cmdFunc func(vmextension.HandlerEnvironment) error
 
 var (
 	cmds = map[string]struct {
-		f             cmdFunc // associated function
-		name          string  // human readable string
-		reportsStatus bool    // determines if running this should log to a .status file
+		f                  cmdFunc // associated function
+		name               string  // human readable string
+		shouldReportStatus bool    // determines if running this should log to a .status file
 	}{
 		"install":   {install, "Install", false},
 		"uninstall": {uninstall, "Uninstall", false},
--- a/cmds.go
+++ b/cmds.go
@@ -1,0 +1,25 @@
+package main
+
+import "github.com/Azure/azure-docker-extension/pkg/vmextension"
+
+type cmdFunc func(vmextension.HandlerEnvironment) error
+
+var (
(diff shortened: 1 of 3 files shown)


cmds.go
```diff
@@ -0,0 +1,25 @@
...
+
+import "github.com/Azure/azure-docker-extension/pkg/vmextension"
+
+type cmdFunc func(vmextension.HandlerEnvironment) error
+
+var (
+	cmds = map[string]struct {
+		f             cmdFunc // associated function
+		name          string  // human readable string
+		reportsStatus bool    // determines if running this should log to a .status file
```
boumenot: shouldReportStatus

cmds.go
```diff
@@ -0,0 +1,25 @@
+package main
+
+import "github.com/Azure/azure-docker-extension/pkg/vmextension"
+
+type cmdFunc func(vmextension.HandlerEnvironment) error
+
+var (
+	cmds = map[string]struct {
+		f             cmdFunc // associated function
```
boumenot: You basically defined a type, but wrapped it up in a map.  Why not just a type?

ahmetalpbalkan: @boumenot map is for lookup with string-->func

cmds.go
ahmetalpbalkan: will fix.

cmds.go
boumenot: Understood.  The compiler can check types, but it cannot do the same for strings.  If you can lean on the compiler to solve this problem I would.

boumenot:
LGTM.

cmds.go
ahmetalpbalkan: @boumenot I miss the full argument. the anonymous struct is still type safe. I just use the map to lookup the user-input command e.g. "enable" to the enable method below.

boumenot: The compiler checks this code.

```go
extension.install()
```

It does not check this code.  You have to do it.

```go
m["install"].f()
```

ahmetalpbalkan: fair, I'm already doing it in this case.

PR #27, opened 2016-07-22
seqnum: update IsSmallerThan→IsSmallerOrEqualThan

ahmetalpbalkan:
This package was designed to be used to make sure we do not
execute same or lower sequence numbers more than once. Changing
the method accordingly to be more useful in that regard.

Signed-off-by: Ahmet Alp Balkan

### DIFF
--- a/seqnum/seqnum.go
+++ b/seqnum/seqnum.go
@@ -24,16 +24,17 @@ func Set(path string, num int) error {
 	return errors.Wrap(ioutil.WriteFile(path, b, chmod), "seqnum: failed to write")
 }
 
-// IsSmallerThan returns if the sequence number stored at path is smaller than
-// the provided num. If no number is stored, returns true and no error.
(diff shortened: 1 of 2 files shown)


seqnum/seqnum.go
```diff
@@ -24,16 +24,17 @@ func Set(path string, num int) error {
 	return errors.Wrap(ioutil.WriteFile(path, b, chmod), "seqnum: failed to write")
 }
 
-// IsSmallerThan returns if the sequence number stored at path is smaller than
-// the provided num. If no number is stored, returns true and no error.
-func IsSmallerThan(path string, num int) (bool, error) {
+// IsSmallerOrEqualThan returns if the sequence number stored at path is smaller
```
paulmey: ... returns _true_ if ...

ahmetalpbalkan: nice catch. will fix

paulmey:
one nit... LGTM

PR #28, opened 2016-07-22
.gitignore minor update

ahmetalpbalkan:
This way we only ignore "bin" at the root level and not in
nested directories. I'll be adding a Makefile to utilize these dirs.

Signed-off-by: Ahmet Alp Balkan <ahmetalpbalkan@gmail.com>


paulmey:
LGTM

PR #30, opened 2016-07-25
Vendor: github.com/Azure/azure-docker-extension

ahmetalpbalkan:
Vendor-only change.

Incorporate latest changes (PRs 96, 97, 98, 99) from azure-docker-extension repo
where we grab extension handler authoring utility pkgs for Go.


ahmetalpbalkan:
Build fails but the next PR will fix it...

vendor/github.com/Azure/azure-docker-extension/pkg/vmextension/handlerenv.go
```diff
@@ -3,8 +3,18 @@ package vmextension
 import (
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
 )
 
+// HandlerEnvFileName is the file name of the Handler Environment as placed by the
```
boumenot: placed => created ?

vendor/github.com/Azure/azure-docker-extension/pkg/vmextension/handlersettings.go
```diff
@@ -9,7 +9,7 @@ import (
 	"io/ioutil"
 	"path/filepath"
 
-	"github.com/Azure/azure-docker-extension/pkg/executil"
+	"os/exec"
```
boumenot: extraneous new line

vendor/github.com/Azure/azure-docker-extension/pkg/vmextension/handlerenv.go
ahmetalpbalkan: meh, both are the same in this case. also if we need to fix it, it'd be better to fix it upstream

boumenot:
LGTM.

vendor/github.com/Azure/azure-docker-extension/pkg/vmextension/handlersettings.go
ahmetalpbalkan: ah fair vscode keeps failing me on imports, but I'll keep it the same as it is harmless for now.

vendor/github.com/Azure/azure-docker-extension/pkg/vmextension/handlerenv.go
boumenot: You're right.  I forgot this was a vendor'ed.

PR #31, opened 2016-07-26
Implement main binary

ahmetalpbalkan:
This change actually gives us a functioning extension handler.

- We store data at /var/lib/azure/custom-script (created upon 'install')
    - This is different than what old extension does.
- This data directory removed upon 'uninstall'.
- Opereations 'update' and 'disable' is no-op (same as before)
- Processed sequence number file (formerly known as 'mrseq') now resides
    at this data directory. This way agent does not need to move the
    'mrseq' file around.
- Operation 'enable' is where the magic happens and it is clear to follow.
    - parses handler settings
    - creates download/output directory
    - download the files
    - runs the command
- There is no 'daemon' subcommand yet as I am planning to invoke 'enable'
  via a shim that does double-fork and log forwarding to a custom log path
  (e.g. /var/log/azure/custom-script.log)
- Moved the check for 'commandToExecute' not empty from json-schema validation
  to logical validation.

Signed-off-by: Ahmet Alp Balkan <ahmetalpbalkan@gmail.com>


cmds.go
```diff
@@ -1,25 +1,122 @@
...
+}
+
+func enable(ctx *log.Context, h vmextension.HandlerEnvironment) error {
+	// parse the extension handler settings (not available prior to 'enable')
+	cfg, err := parseAndValidateSettings(ctx, h.HandlerEnvironment.ConfigFolder)
+	if err != nil {
+		return errors.Wrap(err, "failed to get configuration")
+	}
+
+	seq, err := strconv.Atoi(h.SeqNo)
```
boumenot: Why can't parseAndValidateSettings validate this?

cmds.go
```diff
@@ -1,25 +1,122 @@
...
+	}
+
+	// exit if this sequence number (a snapshot of the configuration) is
+	// processed. if not, save this sequence number before proceeding.
+	sf := filepath.Join(dataDir, seqNumFile)
+	ctx.Log("event", "comparing seqnum")
+	if b, err := seqnum.IsSmallerOrEqualThan(sf, seq); err != nil {
+		return errors.Wrap(err, "failed to check sequence number")
+	} else if b {
+		ctx.Log("event", "exit", "message", "this configuration is already processed, will not run again")
```
boumenot: _this configuration is already..._ sounds less clear to me than "this script" or "this script configuration."  I don't think customers are going to think of it as configuration.  What do you think?

cmds.go
ahmetalpbalkan: hmm it's not part of Settings, it's part of HandlerEnvironment. the issue is we get the seqnum from the handlerenvironment.json as string (if I type it as int, it wouldn't deserialize correctly). This happens in vmextension pkg we get. 

The reason I do Atoi here is so that seqnum.IsSmallerOrEqualsThan can do the comparison.

cmds.go
```diff
@@ -1,25 +1,122 @@
...
+	seq, err := strconv.Atoi(h.SeqNo)
+	if err != nil {
+		return errors.Wrapf(err, "failed to parse seqnum: %q", h.SeqNo)
+	}
+
+	// exit if this sequence number (a snapshot of the configuration) is
+	// processed. if not, save this sequence number before proceeding.
+	sf := filepath.Join(dataDir, seqNumFile)
+	ctx.Log("event", "comparing seqnum")
+	if b, err := seqnum.IsSmallerOrEqualThan(sf, seq); err != nil {
```
boumenot: This if statement feels like a low-level concept mixed into a high-level concept.  The semantic meaning of IsSmallerOrEqualThan means much less than IsConfigProcessed() (<insert better name>).  If you pick a good subroutine name you can remove the comment.

cmds.go
ahmetalpbalkan: @boumenot I completely understand, I adopted this string from the old extension. Will fix.

cmds.go
```diff
@@ -1,25 +1,122 @@
...
+	}
+	if err := ExecCmdInDir(cmd, dir); err != nil {
+		ctx.Log("event", "failed to execute command", "error", err, "output", dir)
+		return errors.Wrap(err, "failed to execute command")
+	}
+	ctx.Log("event", "executed command", "output", dir)
+	ctx.Log("event", "enabled")
+	return nil
+}
+
```
boumenot: 60 lines for a method is too long.

main.go
```diff
@@ -33,4 +86,47 @@ func printUsage() {
...
+//
+// If an error occurs reporting the status, it will be logged and returned.
+func reportStatus(ctx *log.Context, hEnv vmextension.HandlerEnvironment, t status.Type, c cmd, msg string) error {
+	if !c.shouldReportStatus {
+		ctx.Log("status", "not reported for operation (by design)")
+		return nil
+	}
+	s := status.NewStatus(t, c.name, statusMsg(c, t, msg))
+	seq, _ := strconv.Atoi(hEnv.SeqNo)
+	if err := s.Save(hEnv.HandlerEnvironment.StatusFolder, seq); err != nil {
```
boumenot: Is this an atomic operation?

cmds.go
ahmetalpbalkan: hmm sounds alright, but it'll be another method with just 1 line calling this function. I am not sure if changing it will make it a lot less cryptic.

handlersettings.go
```diff
@@ -31,25 +53,71 @@ type protectedSettings struct {
...
+	}
 	return nil
 }
+
+// toJSON converts given in-memory JSON object representation into a JSON object string.
+func toJSON(o map[string]interface{}) (string, error) {
+	if o == nil { // instead of JSON 'null' assume empty object '{}'
+		return "{}", nil
+	}
+	b, err := json.Marshal(o)
```
boumenot: Any value in using MarshalIndent instead?  If it will ever be displayed to the user or dumped into logs it might be.

main.go
ahmetalpbalkan: @boumenot it is not safe against multiple writers, although it opens file for writing with O_TRUNCH. should we make it an atomic move? I did not hear any issues with this in docker-extension so far however waagent might as well call us twice in rapid succession and that might trigger some issue. I have not tested it against that.

cmds.go
ahmetalpbalkan: good feedback. I guess I can extract file download and executing command logic into separate functions by passing them the logger.

handlersettings.go
ahmetalpbalkan: @boumenot hmm this is just an internal operation and we pass the output to the json-schema validator. if we were to log it, it would be escaped with \t and \n s anyway into a single line due to the structured logger we use.

boumenot:
Good stuff.  I look forward to giving this a try.

main.go
boumenot: @ahmetalpbalkan Yes, make it an atomic move.  We had an issue with one of the marketplace templates where a large number of VMs are deployed.  I have a sample template to reproduce the issue reliably.

handlersettings.go
boumenot: As I read the code more I realized it was only used internally.  Disregard.

main.go
ahmetalpbalkan: @boumenot oh I now recall, it was the issue that the file was partially written when the waagent looked at it and it was therefore malformed JSON. I opened a TODO issue https://github.com/Azure/azure-docker-extension/issues/100 to track this.

ahmetalpbalkan:
@boumenot pushed 2 commits addressing CR comments. PTAL

```
commit 380438333efb579bffc1b9c07bc3aa0df65b3769

    cmds: split enable() into several methods, add tests

    extracted logic in enable() to smaller and testable
    methods.

commit deb55b93a11e9e535154700420e984911b22d356

    main: move status methods to status*.go
```

cmds.go
```diff
@@ -1,25 +1,154 @@
...
+		if err := os.RemoveAll(dataDir); err != nil {
+			return errors.Wrap(err, "failed to delete data dir")
+		}
+		ctx.Log("event", "removed data dir")
+	}
+	ctx.Log("event", "uninstalled")
+	return nil
+}
+
+func enable(ctx *log.Context, h vmextension.HandlerEnvironment) error {
```
boumenot: I like this better; it is easier to read.

boumenot:
LGTM.

ahmetalpbalkan:
@boumenot just pushed one more little commit. feel free to take a look... tests should pass now.

PR #32, opened 2016-07-26
integration-test: add test scenarios

ahmetalpbalkan:
The integration tests use `bats` to run the scenarios as bash scripts.

To test the extension handler functionality, we simply:

- build a Docker image using test.Dockerfile
    - copy some files from ` make it look like a `/var/lib/waagent` directory
    - copy extension binary into the container
- remove the `test` container if it exists
- create a Docker container (name: `test`) from image
    - specify which handler subcommand will be invoked (e.g. `fake-waagent
      install`)
- push .settings file and certificate/private key to the container (.crt, .prv)
    - do other things on the container that we need to craft the environment
- start the container
- collect the output from the command execution
- validate using the following:
    - check status code
    - validate output of the command
    - `docker diff test` to validate file changes in the container
    - copy files out of container and validate their contents

### DIFF
--- a/integration-test/test/basic.bats
+++ b/integration-test/test/basic.bats
@@ -1,0 +1,59 @@
+#!/usr/bin/env bats
+
+load test_helper
+
+@test "meta: docker is installed" {
+    run docker version
+    echo "$output">&2
+    [ "$status" -eq 0 ]
+}
+
+@test "meta: can build the test container image" {
+    run build_docker_image
+    echo "$output"
+    [ "$status" -eq 0 ]
+}
+
+@test "meta: can start the test container" {
+    run in_tmp_container fake-waagent
+    echo "$output"
+    [ "$output" = "Usage: /sbin/fake-waagent <handlerCommand>" ]
+    [ "$status" -eq 1 ]
+}
+
+@test "meta: can create vm private/public keys" {
+    run mk_certs
+    echo "$output"
+    [ "$status" -eq 0  ]
+
+    thumbprint="$output"
+    [ -f "$certs_dir/$thumbprint.prv" ]
+    [ -f "$certs_dir/$thumbprint.crt" ]
+}
+
+@test "meta: encrypt a protected settings" {
+    run mk_certs
+    echo "$output"
+    [ "$status" -eq 0  ]
+
+    tp="$output"
+    run encrypt_settings "$tp" "`seq 0 1000`"
+    echo "$output"
+    [ "$status" -eq 0 ]
+    [ -n "$output" ]
+}
+
+@test "meta: can create a .settings json with public/protected config" {
+    tp="$mk_certs"
+    run mk_settings_json '' '{"commandToExecute":"touch /a.txt"}' "$tp"
+    echo "$output"
+    [ "$status" -eq 0 ]
+}
+
+@test "meta: can create a temporary file" {
+    run save_tmp_file "foobar"
+    echo "$output"
+    [ "$status" -eq 0 ]
+    [ -f "$output" ]
+    [[ "$(cat "$output")" == "foobar" ]]
+}
--- a/integration-test/test/handler-commands.bats
+++ b/integration-test/test/handler-commands.bats
@@ -1,0 +1,120 @@
+#!/usr/bin/env bats
+
+load test_helper
+
+setup(){
+    build_docker_image
+}
+
+teardown(){
+    rm -rf "$certs_dir"
+}
+
+@test "handler command: install - creates the data dir" {
+    run in_container fake-waagent install
+    echo "$output"
+    [ "$status" -eq 0 ]
+    [[ "$output" = *event=installed* ]]
+
+    diff="$(container_diff)"
+    echo "$diff"
+    [[ "$diff" = *"A /var/lib/azure/custom-script"* ]]
+}
+
+@test "handler command: enable - can process empty settings, but fails" {
+    mk_container sh -c "fake-waagent install && fake-waagent enable"
+    push_settings '' ''
+
+    run start_container
+    echo "$output"
+    [ "$status" -eq 1 ]
+    [[ "$output" == *"invalid configuration: 'commandToExecute' is not specified"* ]]
+
+     # Validate .status file says enable failed
+     diff="$(container_diff)"; echo "$diff"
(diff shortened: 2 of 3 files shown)


ahmetalpbalkan:
Tests are expected to fail as this is not rebased on top of #31. I still have not plugged these tests in  travis-ci (I'm not sure it'll work as integration tests require a docker daemon). That'll be another PR.

ahmetalpbalkan:
Running these tests looks like this

```
bats integration-test/test
 ✓ meta: docker is installed
 ✓ meta: can build the test container image
 ✓ meta: can start the test container
 ✓ meta: can create vm private/public keys
 ✓ meta: encrypt a protected settings
 ✓ meta: can create a .settings json with public/protected config
 ✓ meta: can create a temporary file
 ✓ handler command: install - creates the data dir
 ✓ handler command: enable - can process empty settings, but fails
 ✓ handler command: enable - validates json schema
 ✓ handler command: enable - captures stdout/stderr into file
 ✓ handler command: enable - doesn't process the same sequence number again
 ✓ handler command: enable - parses protected settings
 ✓ handler command: enable - downloads files
 ✓ handler command: uninstall - deletes the data dir

15 tests, 0 failures
```

integration-test/test/test_helper.bash
```diff
@@ -0,0 +1,126 @@
...
+start_container() {
+    docker start --attach $TEST_CONTAINER
+}
+
+container_diff() {
+    docker diff $TEST_CONTAINER
+}
+
+container_read_file() { # reads the file at container path $1
+    set -eo pipefail
```
boumenot: Why is this scoped to this method?

integration-test/test/basic.bats
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env bats
+
+load test_helper
+
+@test "meta: docker is installed" {
+    run docker version
+    echo "$output">&2
+    [ "$status" -eq 0 ]
```
boumenot: Should you check status before running echo?  Won't that corrupt "run docker version", which I assume is the intent.

integration-test/test/handler-commands.bats
```diff
@@ -0,0 +1,120 @@
...
+}
+
+@test "handler command: enable - can process empty settings, but fails" {
+    mk_container sh -c "fake-waagent install && fake-waagent enable"
+    push_settings '' ''
+
+    run start_container
+    echo "$output"
+    [ "$status" -eq 1 ]
+    [[ "$output" == *"invalid configuration: 'commandToExecute' is not specified"* ]]
```
boumenot: How valuable is it to verify the error message in detail at this stage.  I feel like you have numerous tests to catch this already, so why do it again.  Is error code sufficient?

integration-test/test/handler-commands.bats
```diff
@@ -0,0 +1,120 @@
...
+    [ "$status" -eq 0 ]
+
+    # Validate contents of stdout/stderr files
+    stdout="$(container_read_file /var/lib/azure/custom-script/download/0/stdout)"
+    echo "stdout=$stdout" && [[ "$stdout" = "HelloStdout" ]]
+    stderr="$(container_read_file /var/lib/azure/custom-script/download/0/stderr)"
+    echo "stderr=$stderr" && [[ "$stderr" = "HelloStderr" ]]
+}
+
+@test "handler command: enable - doesn't process the same sequence number again" {
```
boumenot: I like this one.

integration-test/test/handler-commands.bats
```diff
@@ -0,0 +1,120 @@
...
+    diff="$(container_diff)"; echo "$diff"
+    [[ "$diff" == *"A /a.txt"* ]]
+}
+
+@test "handler command: enable - downloads files" {
+    mk_container sh -c "fake-waagent install && fake-waagent enable"
+    # download an external script and run it
+    push_settings '{
+        "fileUris": [
+                "https://gist.github.com/anonymous/8c83af2923ec8dd4a92309594a6c90d7/raw/f26c2cbf68e22d42f703b78f8a4562c5c8e43ba7/script.sh"
```
boumenot: How long do gists last?

integration-test/test/test_helper.bash
```diff
@@ -0,0 +1,126 @@
...
+    docker cp $TEST_CONTAINER:"$1" - | tar x --to-stdout
+} 
+
+mk_certs() { # creates certs/{THUMBPRINT}.(crt|key) files under ./certs/ and prints THUMBPRINT
+    set -eo pipefail
+    mkdir -p "$certs_dir" && cd "$certs_dir" && rm -f "$certs_dir/*"
+    openssl req -x509 -newkey rsa:2048 -keyout key.pem -out cert.pem -days 365 -nodes -batch &>/dev/null
+    thumbprint=$(openssl x509 -in cert.pem -fingerprint -noout| sed 's/.*=//g' | sed 's/://g')
```
boumenot: 😁 The variable is called thumbprint, but the command is fingerprint.  You've been here too long!

boumenot:
LGTM.

integration-test/test/test_helper.bash
ahmetalpbalkan: @boumenot I'm just preserving the terminology of waagent (authored by people stayed here too long as well)

integration-test/test/test_helper.bash
ahmetalpbalkan: I believe setting -e globally has issues with bats https://github.com/sstephenson/bats/issues/171

integration-test/test/basic.bats
ahmetalpbalkan: @boumenot nope `$status` contains the exitcode of the command that is last executed by prefix `run ...` thanks to bats. I type output here because if it fails on the `[ "$status" -eq 0 ]` status line, we'll see the output of the command and tell why it failed.

integration-test/test/handler-commands.bats
ahmetalpbalkan: @boumenot maybe I should reduce to `invalid configuration` but error code is not sufficient as it might have failed due to far too many reasons.

integration-test/test/handler-commands.bats
ahmetalpbalkan: @boumenot until they shut down the gist service (or change its url schema), I guess we can do what you said.

PR #33, opened 2016-07-26
seqnum: convert <= logic to <

ahmetalpbalkan:
Changing the seqnum helper method again to give the following truth table:

(stored,given) => outcome
(none, 0) => true
(0, 0) => false
(0, 1) => true
(1, 0) => false
(1, 1) => false
(1, 2) => true

### DIFF
--- a/seqnum/seqnum.go
+++ b/seqnum/seqnum.go
@@ -24,17 +24,17 @@ func Set(path string, num int) error {
 	return errors.Wrap(ioutil.WriteFile(path, b, chmod), "seqnum: failed to write")
 }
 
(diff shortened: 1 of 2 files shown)


boumenot:
LGTM.

My only suggestion is to have a better than IsSmallerThan, or make the entry point into this a different name and make isSmallerThan private.  

I'm struggling on how to articulate this better.  IsSmallerThan is just <, but your method has much more meaning than that. You need a name to capture that.

ahmetalpbalkan:
will consider that in the future if this grows into a more common utility, thanks. merging...

PR #39, opened 2016-07-26
Shim script for extension handler

### LINKED ISSUE #35: [todo] daemonize 'enable'
I am planning to write a bash script shim that 

1. deamonizes the `enable` command
2. redirects the logs of `enable` command to a static location e.g. `/var/log/azure/custom-script/handler.log`

an example of this is here: https://github.com/Azure/azure-docker-extension/blob/master/scripts/run-in-background.sh

in this case, cmds like install/enable/disable/uninstall wouldn't be logging to this path, but I think we should write a shim for those as well (that doesn't daemonize). 

I want this shim because waagent does not collect child processes (e.g. extensions) logs from their stdout/stderr and expects them to write to a LogFolder specified in HandlerEnvironment.json. 

This is quite against the unix philosophy (processes logging themselves to the files) and it's also not reliable as the go process panics, it won't be able to log the panic. But if we have a shim (written in bash script) we can collect the crash logs as well into a file with a simple `.. | tee -a` redirection.

@boumenot what do you think?

ahmetalpbalkan:
Wrote a bash shim that forwards logs of the handler process
reliably and daemonizes the 'enable' command via `nohup .. &`.

Adopted from https://github.com/Azure/azure-docker-extension/blob/master/scripts/run-in-background.sh and extended to be used for all commands (not just 'enable')

Fixes #35.

------

# example pstree output

`custom-script-extension install` (attached to the process tree):

```
# pstree -ap
bash,1
  `-fake-waagent,79 /sbin/fake-waagent install
      `-custom-script-s,88 ./Extension/bin/custom-script-shim install
          |-custom-script-e,94 /var/lib/waagent/Extension/bin/custom-script-extension install
          `-custom-script-s,91 ./Extension/bin/custom-script-shim install
              `-tee,93 -ia /var/log/azure/custom-script/handler.log
```

`custom-script-extension enable` (detached from the process tree):

```
bash,1
  |-custom-script-e,157 /var/lib/waagent/Extension/bin/custom-script-extension enable
  |   `-sleep,158 100
  |-custom-script-s,138 ./Extension/bin/custom-script-shim enable
  |   `-tee,140 -ia /var/log/azure/custom-script/handler.log
```

cc: @boumenot

### DIFF
--- a/misc/custom-script-shim
+++ b/misc/custom-script-shim
@@ -1,0 +1,74 @@
+#!/bin/bash
+
+set -euo pipefail
+readonly SCRIPT_DIR=$(dirname "$0")
+readonly LOG_DIR="/var/log/azure/custom-script"
+readonly LOG_FILE=handler.log
+readonly HANDLER_BIN="custom-script-extension"
+
+
+# status_file returns the .status file path we are supposed to write
+# by determining the highest sequence number from ./config/*.settings files.
+status_file_path() {
+        # normally we would need to find this config_dir by parsing the
+        # HandlerEnvironment.json, but we are in a bash script here,
+        # so assume it's at ../config/.
+        config_dir=$(readlink -f "${SCRIPT_DIR}/../config")
+        status_dir=$(readlink -f "${SCRIPT_DIR}/../status")
+        config_file=$(ls $config_dir | grep -E ^[0-9]+.settings$ | sort -n | tail -n 1)
+        if [ -f "$config_file" ]; then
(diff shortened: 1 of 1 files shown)


ahmetalpbalkan:
cc: @jasonzio as well, reviews appreciated

boumenot:
LGTM.

ahmetalpbalkan:
merging but I appreciate if there are any comments afterwards.

PR #24, opened 2016-07-21
Containerized integration test environment

ahmetalpbalkan:
Built a containerized environment that looks like an actual VM
and wrote a fake agent which you can invoke the extension binary
through.

Refer to the integration-test/env/README.md for more info.

This helps us test:
- agent binary can handle the directory structure presented by waagent
- agent binary can handle the HandlerEnvironment, log dirs etc
- agent binary decrypts/parses the settings correctly
- agent binary can write status to the right place

By containerizing there will be no residue left on the dev machine
while testing the changes.

PR #40, opened 2016-07-26
travis.yml: add integration tests

ahmetalpbalkan:
:warning::warning: this is expected  to fail until #24 and #32 are merged. once they are in, I will rebase and we should start getting CI running our integration test suite as well.

-----

- change sudo:true so that we can get a Docker engine
    - this will slow down the time to get a test box (a few mins)
      but it should be fine
- add bats in 'install'
- print docker info prior to testing
- execute bats

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,10 +7,15 @@ install:
   - sudo add-apt-repository ppa:duggan/bats --yes
   - sudo apt-get update -qq
   - sudo apt-get install -qq bats
+  - curl -sL https://deb.nodesource.com/setup_6.x | sudo bash -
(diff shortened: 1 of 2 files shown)


boumenot:
LGTM.

ahmetalpbalkan:
rebased

ahmetalpbalkan:
@boumenot w00t it worked. Added some commits to make travis work. PTAL later: 4eacb59...792d491

I discovered a flakiness in one test where `wait-for-enable` exited immediately because the nohup did not yet start the child process fully. will patch `wait-for-enable` by adding an initial 1s delay.

PR #43, opened 2016-07-27
integration-test: add storage tests for blobs as fileUris

ahmetalpbalkan:
When AZURE_STORAGE_ACCOUNT and AZURE_STORAGE_ACCESS_KEY passed,
we now have a scenario that runs by uploading two 512k files to
Azure Blob Storage and configures extension to download it.

We verify by checking test container fs diff to see files are
downloaded and we verify checksum on one of the files.

### DIFF
--- a/integration-test/test/handler-commands.bats
+++ b/integration-test/test/handler-commands.bats
@@ -152,7 +152,7 @@ teardown(){
     [[ "$diff" == *"A /var/lib/azure/custom-script/download/0/$blob2"* ]] # file downloaded
 
     # compare checksum
-    existing=$(cat "$tmp" | md5 -q)
+    existing=$(md5 -q "$tmp")
     echo "Local file checksum: $existing"
     got=$(container_read_file "/var/lib/azure/custom-script/download/0/$blob1" | md5 -q)
     echo "Downloaded file checksum: $got"
--- a/integration-test/test/handler-commands.bats
+++ b/integration-test/test/handler-commands.bats
@@ -108,6 +108,57 @@ teardown(){
     [[ "$diff" == *"A /b.txt"* ]] # created by script.sh
 }
 
+@test "handler command: enable - download files from storage account" {
+    if [[ -z  "$AZURE_STORAGE_ACCOUNT" ]] || [[ -z  "$AZURE_STORAGE_ACCESS_KEY" ]]; then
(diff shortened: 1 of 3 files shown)


integration-test/test/handler-commands.bats
```diff
@@ -108,6 +108,57 @@ teardown(){
...
+    diff="$(container_diff)"; echo "$diff"
+    [[ "$diff" == *"A /var/lib/azure/custom-script/download/0/$blob1"* ]] # file downloaded
+    [[ "$diff" == *"A /var/lib/azure/custom-script/download/0/$blob2"* ]] # file downloaded
+
+    # compare checksum
+    existing=$(cat "$tmp" | md5 -q)
+    echo "Local file checksum: $existing"
+    got=$(container_read_file "/var/lib/azure/custom-script/download/0/$blob1" | md5 -q)
+    echo "Downloaded file checksum: $got"
+    [[ "$existing" == "$got" ]]
```
boumenot: NIT: This is fine.  I thought md5 had a way to verify against a file, so no need to reimplement.

boumenot:
LGTM.

integration-test/test/handler-commands.bats
ahmetalpbalkan: @boumenot `container_read_file` returns the file contents from the container directly (i.e. doesn't save to a file) so `... | md5` is good enough. maybe I can change the cat|md5 in `existing=` to `md5 {file}`

ahmetalpbalkan:
addressed `cat {file}|md5` --> `md5 {file}` issue. merging.

boumenot:
That isn't what I meant.

```text
$ md5sum --help
Usage: md5sum [OPTION]... [FILE]...
Print or check MD5 (128-bit) checksums.

With no FILE, or when FILE is -, read standard input.

  -b, --binary         read in binary mode
  -c, --check          read MD5 sums from the FILEs and check them
      --tag            create a BSD-style checksum
  -t, --text           read in text mode (default)

```

I meant could you use **--check** to verify the MD5 sum.  Commit the check file to source, and use that to verify the MD5 sum.

ahmetalpbalkan:
@boumenot hmm I think it's fine to validate it like this for now. I can get both sums and print them to the screen and just do a bash string comparison. so if it fails, we get to see the values that got compared etc.

PR #44, opened 2016-07-27
integration-test: fix for daemonized 'enable' cmd

ahmetalpbalkan:
Since #39 is adding a shim that daemonizes 'enable' command and
exits and shim is the entrypoint of the container command, Docker
will kill the container when that entrypoint exits (will not wait
background process to finish). This adds a `wait-for-enable` command
that we use in integration tests. It runs inside the test container
and loops indefinitely until the background process exits.

Also by detaching enable cmd from the process tree, we lose ability
to get its exitcode, but we will still keep validating its output so we're good.

Should be merged once #39 is in.

### DIFF
--- a/integration-test/env/wait-for-enable
+++ b/integration-test/env/wait-for-enable
@@ -1,0 +1,20 @@
+#!/bin/bash
+
+# This script exists because custom-script-shim will detach the enable command
+# into background and when that happens the custom-script shim will exit,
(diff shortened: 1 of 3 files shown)


boumenot:
Nice.

LGTM.

ahmetalpbalkan:
@boumenot thanks PTAL at #39 as well.

PR #45, opened 2016-07-27
wait-for-enable: add initial 1s delay

ahmetalpbalkan:
Inside the container when we invoke `nohup [...] &` it immediately
returns and `wait-for-enable` starts checking `ps aux` for extension
binary. However since travis environment is not fast, `wait-for-enable`
exited immediately once before the process started.

Adding a 1s delay while polling for exit of extension process should
solve this problem.

cc: @boumenot

### DIFF
--- a/integration-test/env/wait-for-enable
+++ b/integration-test/env/wait-for-enable
@@ -7,6 +7,7 @@
 
 log(){ echo "[wait] time=$(date --rfc-3339=s | sed 's/ /T/') $@">&2; }
 
+sleep 1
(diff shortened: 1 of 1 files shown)


boumenot:
LGTM.

Sleeps are pernicious.

ahmetalpbalkan:
I'm aware. This will slow down the test execution time but I don't see a way of running `nohup something &` and blocking until `something` comes up before proceeding (especially if `something` exits very fast).

PR #47, opened 2016-07-28
EDP: add `make bundle` and manifest.xml

ahmetalpbalkan:
Extension deployment pipeline requirements.

- Added a manifest.xml (which has a version number that should be
  kept in sync with `./VERSION` file)
- Added a `make bundle` command which should produce a `bundle/custom-script-extension.zip`

cc: @boumenot

### DIFF
--- a/manifest.xml
+++ b/manifest.xml
@@ -1,0 +1,18 @@
+<?xml version='1.0' encoding='utf-8' ?>
+<ExtensionImage xmlns="http://schemas.microsoft.com/windowsazure">
+  <ProviderNameSpace>Microsoft.Azure.Extensions</ProviderNameSpace>
+  <Type>CustomScriptForLinux</Type>
+  <Version>2.0.0</Version>
+  <Label>Microsoft Azure Custom Script Extension for Linux Virtual Machines</Label>
+  <HostingResources>VmRole</HostingResources>
+  <MediaLink></MediaLink>
+  <Description>Microsoft Azure Custom Script Extension for Linux Virtual Machines</Description>
+  <IsInternalExtension>true</IsInternalExtension>
+  <Eula>https://github.com/Azure/custom-script-extension-linux/blob/master/LICENSE</Eula>
+  <PrivacyUri>http://www.microsoft.com/privacystatement/en-us/OnlineServices/Default.aspx</PrivacyUri>
+  <HomepageUri>https://github.com/Azure/custom-script-extension-linux</HomepageUri>
+  <IsJsonExtension>true</IsJsonExtension>
+  <SupportedOS>Linux</SupportedOS>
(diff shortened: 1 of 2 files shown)


manifest.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version='1.0' encoding='utf-8' ?>
+<ExtensionImage xmlns="http://schemas.microsoft.com/windowsazure">
+  <ProviderNameSpace>Microsoft.Azure.Extensions</ProviderNameSpace>
```
boumenot: Can you make this Microsoft.OSTCExtensions for now.  It will make the CI experience easier.

boumenot:
Minor change for now, but LGTM.

manifest.xml
ahmetalpbalkan: @boumenot I don't get it?

boumenot: Change the namespace to Microsoft.OSTCExtensions.

The namespace Microsoft.Azure.Extensions requires the creation of additional configuration data before it can be run through CI.  I would prefer to defer that work for now.

ahmetalpbalkan: I we should probably move away from that namespace. It's cryptic and does not really serve our customers (and hard to manage for us). What do we need to do in EDP to modify this?

boumenot: The big ones are an SPN and a storage account with an extensions container that's publically accessible.

It's fine to move away from it, but you're going to make it harder to upgrade.  autoUpgradeMinorVersion is no longer applicable, so customers do not have a simple upgrade path to the newer versions.  It's fine if you want to be _2.0_, but it will not work if you want to be _1.6_.

ahmetalpbalkan: @boumenot I think we can just publish as a new extension under our subs as 2.0. I realized the migration from 1.6 to 2.0 has to be by changing all the templates, documentation and all the search engine results and will take time. We might as well publish it as a new ext.

boumenot: You'll need to provide the following.  Send it to my internal email.

* SubscriptionID of the publishing subscription.
* Thumbprint of the management certificate for the publishing subscription.
* SPN
 * client_id
 * tenant_id
 * client_secret

You'll need to register the SPN with secret store as well as the management certificate.  Please create ones just for the CI.  Please register the client_secret and management certificate in Jenkins.  (I can help with that.)  We'll need the IDs of these Jenkins secrets.

ahmetalpbalkan: @boumenot we can use the same subscription as `Microsoft.Azure.Extensions.LinuxAsm` or `Microsoft.Azure.Extensions.TestLAD`. Are they already in EDP?

boumenot: No. :disappointed:   Hence, the need for all of the info.

PR #48, opened 2016-07-29
integration-test: remove dependency to gist

ahmetalpbalkan:
Checking in the external test file. The CI is supposed to fail and succeed back again once merged.

This is less dependent on github gist and more on github's URL redirection for raw files and the source tree structure now.


garimakhulbe:
LGTM

PR #49, opened 2016-08-01
Move packages into pkg/, handler to main/

ahmetalpbalkan:
This will make root of the repository a bit more readable hopefully.


boumenot:
Meh.  I could go either way.

LGTM.

PR #53, opened 2016-08-02
Do not rely on SeqNo in HandlerEnvironment.xml

ahmetalpbalkan:
Apparently waagent-2.0.x included a `SeqNo` field in HandlerEnvironment.xml
and 2.1.x removed it (it was not covered by the extensions spec). Other
extensions were using `vmextension.FindSeqNum(...)` method to locate the
highest sequence number.

Switching this handler to use it as well and refactoring a few methods
accordingly.

cc: @boumenot

### DIFF
--- a/main/cmds.go
+++ b/main/cmds.go
@@ -1,9 +1,9 @@
 package main
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
-	"strconv"
 
 	"github.com/Azure/azure-docker-extension/pkg/vmextension"
 	"github.com/Azure/custom-script-extension-linux/pkg/seqnum"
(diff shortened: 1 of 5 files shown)


main/cmds.go
```diff
@@ -1,17 +1,17 @@
...
-	"strconv"
 
 	"github.com/Azure/azure-docker-extension/pkg/vmextension"
 	"github.com/Azure/custom-script-extension-linux/pkg/seqnum"
 	"github.com/go-kit/kit/log"
 	"github.com/pkg/errors"
 )
 
-type cmdFunc func(*log.Context, vmextension.HandlerEnvironment) error
+type cmdFunc func(ctx *log.Context, hEnv vmextension.HandlerEnvironment, seqNum int) error
```
boumenot: It's unfortunate you have to pass in another parameter.  You really want the union of hEnv and seqNum as you had before.  I would consider creating the type you want to hold all of the necessary state.

boumenot:
LGTM.

main/cmds.go
ahmetalpbalkan: Got it, it actually appears like only the `enable` function needs:

- sequence number (to save it to a file to prevent running again)
- configFolder (to locate the .settings files)

from HandlerEnvironment.xml. I might consider creating a smaller type.

PR #55, opened 2016-08-02
Vendor azuresdk

ahmetalpbalkan:
- Vendor azure-sdk-for-go latest.
    - contains a fix with handling of special characters in blob/container names
- integration-test: update test to cover blob name with spaces

cc: @boumenot

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -18,8 +18,8 @@
 		},
 		{
 			"ImportPath": "github.com/Azure/azure-sdk-for-go/storage",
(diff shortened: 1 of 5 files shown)


ahmetalpbalkan:
Feel free to merge this and give it another shot. (Do I need to update <Version>)?

boumenot:
You only need to update Version if you are going to release.  The test version is always _major.minor.1337_, and the test version is always cleaned up before proceeding.

boumenot:
LGTM.

This is a nice fix to have!

PR #57, opened 2016-08-11
Unify version string management in code

ahmetalpbalkan:
- Read `main.Version` from `misc/manifest.xml`, this way we only have
  one place to bump the version number before tagging. Also used
  by Extension Deployment Pipeline.
- Let `govvv` handle GitState/GitCommit/BuildDate etc.

cc: @paulmey since @boumenot is OOF.


.travis.yml
```diff
@@ -12,6 +12,7 @@ install:
   - sudo npm install -g npm
   - sudo npm install -g azure-cli
   - go get -u github.com/golang/lint/golint
+  - go get -u github.com/ahmetalpbalkan/govvv
```
paulmey: shouldn't you also update the actual build script to use `govvv` instead of just `go`?

main/version.go
```diff
@@ -5,33 +5,24 @@ import (
...
+
 }
 
 // DetailedVersionString returns a detailed version string including version
 // number, git commit, build date, source tree state and the go runtime version.
 func DetailedVersionString() string {
-	state := State
-	if state == "" {
-		state = "clean"
-	}
```
paulmey: Test failed: [TestDetailedVersionString](https://travis-ci.org/Azure/custom-script-extension-linux/builds/151601635#L1089)

paulmey:
@ahmetalpbalkan looks like this needs some more fixin'

.travis.yml
ahmetalpbalkan: @paulmey Makefile below uses `govvv`

ahmetalpbalkan:
@paulmey pushed the test fix

.travis.yml
paulmey: Ah yes, completely missed that. Sorry.

paulmey:
LGTM

PR #62, opened 2016-08-30
cmds: delete datadir before installing

ahmetalpbalkan:
Remove dataDir first if exists to cleanup remnants of previous
installations of the handler and captured images as the waagent
currently does not clean up this path.

Bumped version to v2.0.1 for release.

CC: @boumenot

### DIFF
--- a/integration-test/test/handler-commands.bats
+++ b/integration-test/test/handler-commands.bats
@@ -21,6 +21,16 @@ teardown(){
     [[ "$diff" = *"A /var/lib/azure/custom-script"* ]]
 }
 
+@test "handler command: install - cleans up data from the previous installation" {
+    mk_container sh -c "DIR=/var/lib/azure/custom-script; mkdir -p \$DIR && \
+        touch \$DIR/LEFTOVER && \
+        fake-waagent install"
+    run start_container
+    diff="$(container_diff)" && echo "$diff"
+    [[ "$diff" != *LEFTOVER* ]]
+}
+
+
 @test "handler command: enable - can process empty settings, but fails" {
     mk_container sh -c "fake-waagent install && fake-waagent enable && wait-for-enable"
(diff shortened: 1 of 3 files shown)


integration-test/test/handler-commands.bats
```diff
@@ -21,6 +21,16 @@ teardown(){
     [[ "$diff" = *"A /var/lib/azure/custom-script"* ]]
```
boumenot: Have you considered moving this to under/var/lib/waagent/<something>/custom-script, or a file under /var/lib/waagent/custom-script.<something>.

boumenot:
It still doesn't fix the issue of remnant propagating to new images.

integration-test/test/handler-commands.bats
ahmetalpbalkan: @boumenot ok this patch basically keeps the path same but deletes it first for images who are captured with remnants of the old handler. guarantee here is "install is executed only once".

what you're suggesting is essentially changing the path altogether. the way I see we execute it is:  in the "enable" command we have a migration code, such as:

    mv -rf /var/lib/azure/custom-script /var/lib/waagent/custom-script

When the hotfix landed to existing VMs, it will move the state to `/var/lib/waagent/custom-script` next time it got enabled (VM reboot or goal state update etc). Then we'd start using this path... For clean VMs, the migration would be a noop.

Do you wanna do this?

boumenot: Yes.

ahmetalpbalkan:
> It still doesn't fix the issue of remnant propagating to new images.

I see, I guess for that I will add some code doing a migration that shells out to `mv -rf`. I'm not planning to implement a recursive file copy in golang for now as the volume of versions out there that would get migrated is still low.

PR #64, opened 2016-09-14
string to array, log path update

neilpeterson:


ahmetalpbalkan:
Thanks @neilpeterson.

PR #71, opened 2016-10-11
Do not overwrite .status file in new enables

ahmetalpbalkan:
Updating custom-script-shim to not to overwrite a placeholder
"transitioning" `.status` file if it already exists.

Updating handler code writing "transitioning" `.status` file to
not to write the `.status` file until the sequence number check
is completed in the "enable" cmd.

Added integration tests to validate this case.

cc: @boumenot

### DIFF
--- a/misc/custom-script-shim
+++ b/misc/custom-script-shim
@@ -25,25 +25,29 @@ status_file_path() {
 }
 
 write_status() {
-	timestamp="$(date --utc --iso-8601=seconds)"
 	status_file="$(status_file_path)"
-	echo "Writing status to $status_file."
-	cat > "$status_file" <<- EOF
-		[
-			{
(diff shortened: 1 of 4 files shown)


misc/custom-script-shim
```diff
@@ -25,25 +25,29 @@ status_file_path() {
...
-				"status": {
-					"operation": "Enable",
-					"status": "transitioning",
-					"formattedMessage": {
-						"lang": "en",
-						"message": "Enable in progress"
+	if [ -f "$status_file" ]; then
+		echo "Not writing a placeholder status file, already exists: $status_file"
+    else
+		echo "Writing a placeholder status file indicating progress before forking: $status_file"
```
boumenot: Indentation is off.

PR #72, opened 2016-10-11
fix tests with new storage account name

ahmetalpbalkan:
The storage tests started to get "no such error" instead of HTTP 403
because the owner of 'accountname' account deleted the account. So
changing it to 'example' for the time being.

Also disabling CLI telemetry prompt, saves 30s in tests.

### DIFF
--- a/pkg/download/blob_test.go
+++ b/pkg/download/blob_test.go
@@ -64,7 +64,7 @@ func Test_blobDownload_getURL(t *testing.T) {
 }
 
 func Test_blobDownload_fails_badCreds(t *testing.T) {
-	d := NewBlobDownload("accountname", "Zm9vCg==", blobutil.AzureBlobRef{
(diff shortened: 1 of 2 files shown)


pkg/download/blob_test.go
```diff
@@ -64,7 +64,7 @@ func Test_blobDownload_getURL(t *testing.T) {
 }
 
 func Test_blobDownload_fails_badCreds(t *testing.T) {
-	d := NewBlobDownload("accountname", "Zm9vCg==", blobutil.AzureBlobRef{
+	d := NewBlobDownload("example", "Zm9vCg==", blobutil.AzureBlobRef{
```
boumenot: Are you just going to this issue again?

ahmetalpbalkan: Yes it may hit this again but it's a real-world test actually making a call to storage, so I tend to keep it for now.

PR #75, opened 2016-11-10
Add PowerShell Scripts and Examples; Fix Log and Script Paths

dstolts:
Added PowerShell Scripts and Examples
Fixed path location of log files
Fixed path location of downloaded script files

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,27 +113,23 @@ For **Azure CLI**, create a `public.json` (and optionally `protected.json`) and 
 	    --private-config-path protected.json
 
 
-``` 
-For PowerShell with variables
 
-PowerShell Variable Help...
-``` 
-	 #You can find values for your variables using various calls such as: 
-	 #Find VmName : Get-AzureRmVm -ResourceGroupName $rgName
-	 #Find StorageAccount : Get-AzureRMStorageAccount  -ResourceGroupName $rgName
-	 # Find StorageAccountKey:	 
-	 #   $StorageAccount = Get-AzureRmStorageAccount -StorageAccountName $StorageAccountName  -ResourceGroupName $rgName # Get the Storage Account 
-	 #   $StorageAccountKey = (Get-AzureRmStorageAccountKey -ResourceGroupName $rgName -StorageAccountName $StorageAccountName).Value[0] # Get the primary Key 
-     #   $StorageAccountContext = New-AzureStorageContext -StorageAccountKey $StorageAccountKey -StorageAccountName $StorageAccountName # Get the Context 
-	 #   $UriEndpoint = $StorageAccountContext.BlobEndPoint  # EndPoint URL
-	 # Grab the endpoint URL
-	 #   $filename = "hello.sh"
-	 #   $ConfigScriptUri = $StorageAccountContext.BlobEndPoint + "<container>/"+$fileName
-     #   $ConfigScriptFileName =  $fileName
+_For PowerShell with variables_
 
-``` 
+ *You can find values for your variables using various calls such as: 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -91,14 +115,99 @@ For **Azure CLI**, create a `public.json` (and optionally `protected.json`) and
...
+  
+```
+
 # 3. Troubleshooting
 
-Your files are downloaded to a path like: `/var/lib/waagent/custom-script/download/0/` and
+Your files are downloaded to a path like: 
+   `/var/lib/waagent/<Publisher>.<ExtensionName>-<version>/#/ScriptName.ext` 
+    Example: 
+	  `/var/lib/waagent/Microsoft.OSTCExtensions.CustomScriptForLinux-1.5.2.1/download/0/hello.sh`
```
ahmetalpbalkan: This path doesn't sound correct. You're using the old extension at https://github.com/Azure/azure-linux-extensions repo :)

README.md
```diff
@@ -91,14 +115,99 @@ For **Azure CLI**, create a `public.json` (and optionally `protected.json`) and
 
 
 
+_For PowerShell with variables_
```
ahmetalpbalkan: This looks way too detailed to me. Can't we just leave it to the reader as an exercise to extract reusable details into a variables and make this command shorter?

README.md
```diff
@@ -91,14 +115,99 @@ For **Azure CLI**, create a `public.json` (and optionally `protected.json`) and
...
+ All PowerShell examples assume you are successfully connected to an Azure subscription. 
+   For more information see: http://itproguru.com/expert/2016/04/powershell-working-with-azure-resource-manager-rm-step-by-step-changing-rm-subscriptions/
```
ahmetalpbalkan: I'm not sure we're cleared to endorse 3rd party tutorials cc: @neilpeterson what do you think?

README.md
```diff
@@ -91,14 +115,99 @@ For **Azure CLI**, create a `public.json` (and optionally `protected.json`) and
...
+Write-Host "     /var/log/azure/$Publisher.$ExtensionName/$version/CommandExecution.log " -ForegroundColor Green
+Write-Host "     /var/log/azure/$Publisher.$ExtensionName/$version/extension.log" -ForegroundColor Green
+  
+```
+
 # 3. Troubleshooting
 
-Your files are downloaded to a path like: `/var/lib/waagent/custom-script/download/0/` and
+Your files are downloaded to a path like: 
+   `/var/lib/waagent/<Publisher>.<ExtensionName>-<version>/#/ScriptName.ext`
```
ahmetalpbalkan: this is not correct. The path you deleted is the correct one for `Microsoft.Azure.Extensions.CustomScript`, which is hosted in this repo.

README.md
neilpeterson: Preference would be to use this document - [How to install and configure Azure PowerShell](https://azure.microsoft.com/en-us/documentation/articles/powershell-install-configure/).

Thanks

ahmetalpbalkan:
@dstolts Hello again, sorry for the delay. I am still inclined to reject this patch, simply because it goes into far too much detailed usage with PowerShell and makes the API.

We can't afford explaining CLI/PS/REST/ARMTemplate all in a brief `README.md`. To provide a good documentation experience, we host separate articles on https://azure.microsoft.com/documentation/ that accomplish the same task using Azure CLI and PowerShell, on separate pages, with an easy navigation to switch between them.

I think your patches would be more welcomed to the Azure docs repository, which is also open source.

dstolts:
Ok thanks.
I still plan on simplifying for your site.   Just no cycles at the moment.

Issue #67, opened 2016-09-23
Expose stdout/stderr of the script in status file (and on portal/resource explorer)


xpillons:
With the OSTCExtensions release, stdout and stderr were displayed into the status of the instance view of a VM. In my case the message is "Enable in progress" and displayStatus is "Transitioning" while the script is finished and well executed.

ahmetalpbalkan:
@xpillons thanks for reporting. We omitted showing script output on the API/portal/resource explorer on the new version. However the "Transitioning" status should be going away once the script is finished.

Can you please take a screenshot from the portal and grab the files in `/var/lib/waagent/Microsoft.Azure.Extensions.CustomScript-2.0.1/status/*` and attach them here for troubleshooting? Having the `/var/log/azure/custom-script/handler.log` would also help.

xpillons:
Sure, will do it Monday as I’ve tear down the machines now.

xpillons:
Here is the content of the file 0-status and the handler.log files
[cid:image001.png@01D2180D.2E683B30]

Thanks,
Xavier

ahmetalpbalkan:
@xpillons Please reply on GitHub.

xpillons:
Here is the handler content

+ /var/lib/waagent/Microsoft.Azure.Extensions.CustomScript-2.0.1/bin/custom-script-extension install
time=2016-09-26T13:30:21Z version=v2.0.1/git@0f92134-clean operation=install seq=0 event=start
time=2016-09-26T13:30:21Z version=v2.0.1/git@0f92134-clean operation=install seq=0 status="not reported for operation (by design)"
time=2016-09-26T13:30:21Z version=v2.0.1/git@0f92134-clean operation=install seq=0 event="created data dir" path=/var/lib/waagent/custom-script
time=2016-09-26T13:30:21Z version=v2.0.1/git@0f92134-clean operation=install seq=0 event=installed
time=2016-09-26T13:30:21Z version=v2.0.1/git@0f92134-clean operation=install seq=0 status="not reported for operation (by design)"
time=2016-09-26T13:30:21Z version=v2.0.1/git@0f92134-clean operation=install seq=0 event=end
Writing status to /var/lib/waagent/Microsoft.Azure.Extensions.CustomScript-2.0.1/status/0.status.
+ nohup /var/lib/waagent/Microsoft.Azure.Extensions.CustomScript-2.0.1/bin/custom-script-extension enable
[...]

PR #78, opened 2016-11-28
Tail stdout/stderr files in enable status message

ahmetalpbalkan:
See #67 for more info.

This change adds executed tail (max: 4k) of the executed script's stderr/stdout into the status message delivered via `.status` file. A new status file is in the format of:

    Enable (succeeded|failed): \n[stdout]\n(..bytes..)\n[stderr]\n(..bytes..)

And it looks like this in practice where `"commandToExecute": "echo HelloStdout>&1; echo HelloStderr>&2"`:

```json
[
  {
    "version": 1,
    "timestampUTC": "2016-11-28T23:41:31Z",
    "status": {
      "operation": "Enable",
      "status": "success",
      "formattedMessage": {
        "lang": "en",
        "message": "Enable succeeded: \n[stdout]\nHelloStdout\n\n[stderr]\nHelloStderr\n"
      }
    }
  }
]
```

cc: @boumenot

### DIFF
--- a/main/cmds.go
+++ b/main/cmds.go
@@ -11,7 +11,7 @@ import (
 	"github.com/pkg/errors"
 )
 
-type cmdFunc func(ctx *log.Context, hEnv vmextension.HandlerEnvironment, seqNum int) error
+type cmdFunc func(ctx *log.Context, hEnv vmextension.HandlerEnvironment, seqNum int) (msg string, err error)
 type preFunc func(ctx *log.Context, seqNum int) error
 
 type cmd struct {
@@ -20,6 +20,10 @@ type cmd struct {
 	shouldReportStatus bool    // determines if running this should log to a .status file
 	pre                preFunc // executed before any status is reported
 }
+
+const (
+	maxTailLen = 4 * 1024 // length of max stdout/stderr to be transmitted in .status file
+)
 
 var (
 	cmdInstall   = cmd{install, "Install", false, nil}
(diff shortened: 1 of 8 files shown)


boumenot:
LGTM, minus the failing test cases.  From the failure signature, it looks like an easy fix.

main/cmds.go
```diff
@@ -82,24 +86,39 @@ func enablePre(ctx *log.Context, seqNum int) error {
...
+	stdoutF, stderrF := logPaths(dir)
+	stdoutTail, err := tailFile(stdoutF, maxTailLen)
+	if err != nil {
+		ctx.Log("message", "error tailing stdout logs", "error", err)
+	}
+	stderrTail, err := tailFile(stderrF, maxTailLen)
+	if err != nil {
+		ctx.Log("message", "error tailing stderr logs", "error", err)
 	}
+	msg := fmt.Sprintf("\n[stdout]\n%s\n[stderr]\n%s", string(stdoutTail), string(stderrTail))
```
ahmetalpbalkan: There might be people relying on that in other places but I just did not see any of those in ARM templates. ARM templates mostly rely on splitting by \n. I feel like this would also be a good reminder that they shouldn't rely on undocumented features when users get broken downstream and have to fix it?

ahmetalpbalkan:
Just bumped to go1.7 as `io.SeekEnd` is introduced in go1.7 and `os.SEEK_END` is now deprecated.

main/cmds.go
boumenot: Let's see what the impact of the change is.  I am OK with it.

marrobi:
When will this be released? Thanks.

PR #111, opened 2017-10-13
support script command

boumenot:
* report stdout/stderr on failure
 * support fileUris in protected settings
 * telemetry
 * allow user to turn of dos2unix conversion
 * better exit codes to track failure cases


boumenot:
The roll-out will start today, and should complete within a week barring any issues.

PR #123, opened 2017-11-22
Find bash from env

jbgi:
So that custom-script-shim works on distros where bash is not in /bin

### DIFF
--- a/misc/custom-script-shim
+++ b/misc/custom-script-shim
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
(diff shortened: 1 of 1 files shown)


boumenot:
LGTM.

jbgi:
@boumenot any idea when this simple fix can be released?

boumenot:
I do not have an ETA.  This change isn't for an [endorsed distro](https://docs.microsoft.com/en-us/azure/virtual-machines/linux/endorsed-distros), so it will go out when there is a change that affects endorsed too.

PR #145, opened 2018-12-10
Removing stdout/stderr from telemetry

Diastro:
Removing stdout/stderr from telemetry

### DIFF
--- a/main/cmds.go
+++ b/main/cmds.go
@@ -169,22 +169,16 @@ func enable(ctx *log.Context, h vmextension.HandlerEnvironment, seqNum int) (str
 		ctx.Log("message", "error tailing stderr logs", "error", err)
 	}
 
-	msg := fmt.Sprintf("\n[stdout]\n%s\n[stderr]\n%s", string(stdoutTail), string(stderrTail))
-
-	minStdout := min(len(stdoutTail), maxTelemetryTailLen)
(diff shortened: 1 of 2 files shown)


main/cmds.go
```diff
@@ -169,22 +169,16 @@ func enable(ctx *log.Context, h vmextension.HandlerEnvironment, seqNum int) (str
...
-
-	minStdout := min(len(stdoutTail), maxTelemetryTailLen)
-	minStderr := min(len(stderrTail), maxTelemetryTailLen)
-	msgTelemetry := fmt.Sprintf("\n[stdout]\n%s\n[stderr]\n%s",
-		string(stdoutTail[len(stdoutTail)-minStdout:]),
-		string(stderrTail[len(stderrTail)-minStderr:]))
-
 	isSuccess := runErr == nil
-	telemetry("Output", msgTelemetry, isSuccess, 0)
+	telemetry("Output", "-- stdout/stderr omitted from telemetry pipeline --", isSuccess, 0)
```
boumenot: I wouldn't event bother with this message because it is just a needless message.

I am surprised there isn't a corresponding test change, but I could be misremembering.

Diastro: My thought process was just in case someone was expecting to see the stdout in the logs, seeing nothing could be misleading in thinking that nothing was actually returned.
For the test, I re-ran all the tests everything passed. Were you thinking of unit test or some functional ones?

PR #149, opened 2019-06-07
Hide URLs in error messages

D1v38om83r:
Hide URLs in error messages. Status might give away SAS tokens.


pkg/urlutil/urlerrorhelper.go
```diff
@@ -0,0 +1,26 @@
...
+	"net/url"
+	"strings"
+)
+
+func RemoveUrlFromErr(err error) error {
+	strSegments := strings.Split(err.Error(), " ")
+	for i, v := range strSegments {
+		if IsValidUrl(v) {
+			// we found a url
+			strSegments[i] = "[REDACTED]"
```
nkuchta: I think there is still value in logging and including portions of the URL especially for diagnostic/debugging. Can you redact just the query parameters in the URL string?

D1v38om83r: I cannot redact what I cannot parse. So, a malformed url, say,  http// could pass through to the error in the end. So I catch that possibility here. The error here is wrapped with failed to download file[%d], index, so we'll still know which of the provided urls failed.

D1v38om83r: This behavior is consistent with other errors like getting a 404. But there error prints the index of the url provided that failed, so we can still know which one failed.

pkg/urlutil/urlerrorhelper_test.go
```diff
@@ -0,0 +1,43 @@
...
+	"strings"
+	"testing"
+)
+
+func Test_urlparse01(t *testing.T) {
+	// error string as expected
+	errorString := `Get https://aaaaaaaa.blob.core.windows.net/nodeagentpackage-version9-0-0-381/Ubuntu-16.04/batch_config-ubuntu-16.04-1.5.9.tar.gz?sv=2018-03-28&sr=b&sig=a%secret%2Bsecret&st=2019-05-17T01%3A25%3A42Z&se=2021-05-24T01%3A25%3A42Z&sp=r: dial tcp 13.68.165.64:443: i/o timeout`
+	inputErr := fmt.Errorf("%s", errorString)
+	outputErr := RemoveUrlFromErr(inputErr)
+	if strings.Contains(outputErr.Error(), "https://") || !strings.Contains(outputErr.Error(), "[REDACTED]") {
```
Diastro: Might be good to assert that the secret value isn't present. A change could be introduced for example where [redacted] is present but the secret is also, and we should catch this.

D1v38om83r: done

pkg/urlutil/urlerrorhelper_test.go
```diff
@@ -0,0 +1,43 @@
...
+	"strings"
+	"testing"
+)
+
+func Test_urlparse01(t *testing.T) {
+	// error string as expected
+	errorString := `Get https://aaaaaaaa.blob.core.windows.net/nodeagentpackage-version9-0-0-381/Ubuntu-16.04/batch_config-ubuntu-16.04-1.5.9.tar.gz?sv=2018-03-28&sr=b&sig=a%secret%2Bsecret&st=2019-05-17T01%3A25%3A42Z&se=2021-05-24T01%3A25%3A42Z&sp=r: dial tcp 13.68.165.64:443: i/o timeout`
+	inputErr := fmt.Errorf("%s", errorString)
+	outputErr := RemoveUrlFromErr(inputErr)
+	if strings.Contains(outputErr.Error(), "https://") || strings.Contains(outputErr.Error(), "secret") || !strings.Contains(outputErr.Error(), "[REDACTED]") {
```
jscalev: Why not just compare the errorString to the exact one you expect?

pkg/urlutil/urlerrorhelper_test.go
```diff
@@ -0,0 +1,43 @@
+package urlutil
+
+import (
+	"fmt"
+	"strings"
+	"testing"
+)
+
+func Test_urlparse01(t *testing.T) {
```
jscalev: please name the tests something other than Test_urlparse01, etc. For instance, this one could be Test_UrlRedacted or something else that lets us know what's being verified

pkg/urlutil/urlerrorhelper_test.go
```diff
@@ -0,0 +1,43 @@
...
+		fmt.Println(outputErr.Error())
+	}
+}
+
+func Test_urlparse02(t *testing.T) {
+	// error string where scheme is not https because missing space between Get and https://
+	errorString := `Gethttps://aaaaaaaa.blob.core.windows.net/nodeagentpackage-version9-0-0-381/Ubuntu-16.04/batch_config-ubuntu-16.04-1.5.9.tar.gz?sv=2018-03-28&sr=b&sig=a%secret%2Bsecret&st=2019-05-17T01%3A25%3A42Z&se=2021-05-24T01%3A25%3A42Z&sp=r: dial tcp 13.68.165.64:443: i/o timeout`
+	inputErr := fmt.Errorf("%s", errorString)
+	outputErr := RemoveUrlFromErr(inputErr)
+	if strings.Contains(outputErr.Error(), "https://") || strings.Contains(outputErr.Error(), "secret") || !strings.Contains(outputErr.Error(), "[REDACTED]") {
```
jscalev: but "dial tcp 13.68.165.64:443: i/o timeout" should still be there, right? Another reason to just compare the entire string with what we expect.

pkg/urlutil/urlerrorhelper.go
```diff
@@ -0,0 +1,26 @@
...
+		}
+	}
+	return fmt.Errorf(strings.Join(strSegments, " "))
+}
+
+func IsValidUrl(urlstring string) bool {
+	u, parseError := url.Parse(urlstring)
+	if parseError == nil && u.Scheme != "" && u.Host != "" && u.Path != "" {
```
jscalev: what if there's no path? - https://agent.fsb.ru/?sekret=12345

pkg/urlutil/urlerrorhelper_test.go
```diff
@@ -0,0 +1,43 @@
...
+	// error string where the uri isn't separated by space with the rest of the error message
+	errorString := `Gethttps://aaaaaaaa.blob.core.windows.net/nodeagentpackage-version9-0-0-381/Ubuntu-16.04/batch_config-ubuntu-16.04-1.5.9.tar.gz?sv=2018-03-28&sr=b&sig=a%secret%2Bsecret&st=2019-05-17T01%3A25%3A42Z&se=2021-05-24T01%3A25%3A42Z&sp=r:dial tcp 13.68.165.64:443:i/o timeout`
+	inputErr := fmt.Errorf("%s", errorString)
+	outputErr := RemoveUrlFromErr(inputErr)
+	if strings.Contains(outputErr.Error(), "https://") || strings.Contains(outputErr.Error(), "secret") || !strings.Contains(outputErr.Error(), "[REDACTED]") {
+		t.Error("Url removal failed")
+	} else {
+		fmt.Println(outputErr.Error())
+	}
+}
```
jscalev: more possible tests
- url is entire error string (make sure there are no null refs)
- url = "                  " (lots of spaces)

pkg/urlutil/urlerrorhelper.go
D1v38om83r: Good point, I had overlooked that possibility.

pkg/urlutil/urlerrorhelper_test.go
D1v38om83r: done

pkg/urlutil/urlerrorhelper_test.go
D1v38om83r: added if condition for that

pkg/urlutil/urlerrorhelper_test.go
D1v38om83r: added more tests

pkg/urlutil/urlerrorhelper_test.go
D1v38om83r: the error would be less descriptive in describing what exactly went wrong, it the removal failed or it removed non-url parts.

PR #151, opened 2019-07-26
RBAC support for downloading scripts with managed identities

D1v38om83r:
Enable CSE linux to use managed identities to access azure storage blobs to download scripts.

### DIFF
--- a/main/handlersettings.go
+++ b/main/handlersettings.go
@@ -91,11 +91,11 @@ type publicSettings struct {
 // protectedSettings is the type decoded and deserialized from protected
 // configuration section. This should be in sync with protectedSettingsSchema.
 type protectedSettings struct {
-	CommandToExecute       string           `json:"commandToExecute"`
-	Script                 string           `json:"script"`
-	FileURLs               []string         `json:"fileUris"`
-	StorageAccountName     string           `json:"storageAccountName"`
-	StorageAccountKey      string           `json:"storageAccountKey"`
+	CommandToExecute       string            `json:"commandToExecute"`
+	Script                 string            `json:"script"`
+	FileURLs               []string          `json:"fileUris"`
+	StorageAccountName     string            `json:"storageAccountName"`
+	StorageAccountKey      string            `json:"storageAccountKey"`
 	ManagedServiceIdentity *clientOrObjectId `json:managedServiceIdentity`
 }
 
--- a/main/handlersettings.go
+++ b/main/handlersettings.go
@@ -9,11 +9,13 @@ import (
 )
 
 var (
-	errStoragePartialCredentials = errors.New("both 'storageAccountName' and 'storageAccountKey' must be specified")
-	errCmdTooMany                = errors.New("'commandToExecute' was specified both in public and protected settings; it must be specified only once")
-	errScriptTooMany             = errors.New("'script' was specified both in public and protected settings; it must be specified only once")
-	errCmdAndScript              = errors.New("'commandToExecute' and 'script' were both specified, but only one is validate at a time")
-	errCmdMissing                = errors.New("'commandToExecute' is not specified")
+	errStoragePartialCredentials    = errors.New("both 'storageAccountName' and 'storageAccountKey' must be specified")
+	errCmdTooMany                   = errors.New("'commandToExecute' was specified both in public and protected settings; it must be specified only once")
+	errScriptTooMany                = errors.New("'script' was specified both in public and protected settings; it must be specified only once")
+	errCmdAndScript                 = errors.New("'commandToExecute' and 'script' were both specified, but only one is validate at a time")
+	errCmdMissing                   = errors.New("'commandToExecute' is not specified")
+	errUsingBothKeyAndMsi           = errors.New("'storageAccountName' or 'storageAccountKey' must not be specified with 'managedServiceIdentity'")
+	errUsingBothClientIdAndObjectId = errors.New("only one of 'clientId' or 'objectId' must be specified with 'managedServiceIdentity'")
 )
 
 // handlerSettings holds the configuration of the extension handler.
@@ -66,6 +68,14 @@ func (h handlerSettings) validate() error {
 		return errStoragePartialCredentials
 	}
 
+	if (h.protectedSettings.StorageAccountKey != "" || h.protectedSettings.StorageAccountName != "") && !h.protectedSettings.ManagedServiceIdentity.isEmpty() {
+		return errUsingBothKeyAndMsi
+	}
+
+	if h.protectedSettings.ManagedServiceIdentity.ClientId != "" && h.protectedSettings.ManagedServiceIdentity.ObjectId != "" {
+
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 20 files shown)


pkg/download/blobwithmsitoken.go
```diff
@@ -0,0 +1,98 @@
...
+	return &blobWithMsiToken{url, msiProvider}
+}
+
+func GetMsiProviderForStorageAccountsImplicitly(blobUri string) MsiProvider {
+	msiProvider := msi.NewMsiProvider(httputil.NewSecureHttpClient(httputil.DefaultRetryBehavior))
+	return func() (msi.Msi, error) { return msiProvider.GetMsiForResource(GetResourceNameFromBlobUri(blobUri)) }
+}
+
+func GetMsiProviderForStorageAccountsWithClientId(blobUri, clientId string) MsiProvider {
+	msiProvider := msi.NewMsiProvider(httputil.NewSecureHttpClient(httputil.DefaultRetryBehavior))
```
vivlingaiah: should we check clientId is valid here ?

pkg/download/blobwithmsitoken.go
```diff
@@ -0,0 +1,98 @@
...
+	return func() (msi.Msi, error) {
+		return msiProvider.GetMsiUsingClientId(clientId, GetResourceNameFromBlobUri(blobUri))
+	}
+}
+
+func GetMsiProviderForStorageAccountsWithObjectId(blobUri, objectId string) MsiProvider {
+	msiProvider := msi.NewMsiProvider(httputil.NewSecureHttpClient(httputil.DefaultRetryBehavior))
+	return func() (msi.Msi, error) {
```
vivlingaiah: should we check objectid is valid here ?

main/handlersettings.go
```diff
@@ -66,6 +68,16 @@ func (h handlerSettings) validate() error {
 		return errStoragePartialCredentials
 	}
 
+	if (h.protectedSettings.StorageAccountKey != "" || h.protectedSettings.StorageAccountName != "") && !h.protectedSettings.ManagedIdentity.isEmpty() {
```
vivlingaiah: do we need a null check for h.protectedSettings.ManagedIdentity ?

main/schema.go
```diff
@@ -73,7 +73,23 @@ const (
       "description": "Key for the Azure Storage Account (a base64 encoded string)",
       "type": "string",
       "pattern": "^(?:[A-Za-z0-9+/]{4})*(?:[A-Za-z0-9+/]{2}==|[A-Za-z0-9+/]{3}=|[A-Za-z0-9+/]{4})$"
-    }
+    },
+	"managedIdentity": {
```
vivlingaiah: correct indent

pkg/download/blobwithmsitoken.go
D1v38om83r: Already checked as a part of json schema check https://github.com/Azure/custom-script-extension-linux/blob/10710fb27d7fb5061d9193f3f6edbb8b82b1a5a8/main/schema.go#L89

pkg/download/blobwithmsitoken.go
D1v38om83r: Already checked as a part of JSON schema check

https://github.com/Azure/custom-script-extension-linux/blob/10710fb27d7fb5061d9193f3f6edbb8b82b1a5a8/main/schema.go#L84

main/handlersettings.go
D1v38om83r: Good catch, that was definitely a bug

pkg/download/blobwithmsitoken.go
vivlingaiah: I meant whether clientId has permissions

main/schema.go
D1v38om83r: fixed in new commit

pkg/download/blobwithmsitoken.go
D1v38om83r: The only way to figure that is to actually attempt the download. If it doesn't have permissions, the download will fail and the set-extensions call itself with fail

pkg/download/blobwithmsitoken.go
```diff
@@ -57,20 +57,38 @@ func NewBlobWithMsiDownload(url string, msiProvider MsiProvider) Downloader {
 
 func GetMsiProviderForStorageAccountsImplicitly(blobUri string) MsiProvider {
 	msiProvider := msi.NewMsiProvider(httputil.NewSecureHttpClient(httputil.DefaultRetryBehavior))
-	return func() (msi.Msi, error) { return msiProvider.GetMsiForResource(GetResourceNameFromBlobUri(blobUri)) }
+	return func() (msi.Msi, error) {
+		msi, err := msiProvider.GetMsiForResource(GetResourceNameFromBlobUri(blobUri))
+		if err != nil {
+			return msi, errors.Wrapf(err, "Unable to get managed identity. "+
+				"Please make sure that system assigned managed identity is enabled on the VM"+
```
vivlingaiah: add space in end "enabled on the VM "

pkg/download/downloader.go
```diff
@@ -52,5 +52,13 @@ func Download(d Downloader) (int, io.ReadCloser, error) {
 	if resp.StatusCode == http.StatusOK {
 		return resp.StatusCode, resp.Body, nil
 	}
-	return resp.StatusCode, nil, fmt.Errorf("unexpected status code: got=%d expected=%d", resp.StatusCode, http.StatusOK)
+
+	err = fmt.Errorf("unexpected status code: got=%d expected=%d", resp.StatusCode, http.StatusOK)
```
vivlingaiah: use "actual" instead of "got"

pkg/download/downloader.go
```diff
@@ -52,5 +52,13 @@ func Download(d Downloader) (int, io.ReadCloser, error) {
...
+	switch d.(type) {
+	case *blobWithMsiToken:
+		if resp.StatusCode == http.StatusForbidden {
```
vivlingaiah: do you get unauthorized (http 401) in any case ?

pkg/download/blobwithmsitoken.go
D1v38om83r: fixed

pkg/download/downloader.go
D1v38om83r: We only get 404's

pkg/download/downloader.go
D1v38om83r: fixed

pkg/download/blobwithmsitoken.go
```diff
@@ -0,0 +1,116 @@
...
+)
+
+const (
+	xMsVersionHeaderName = "x-ms-version"
+	xMsVersionValue      = "2018-03-28"
+	storageResourceName  = "https://storage.azure.com/"
+)
+
+var azureBlobDomains = map[string]interface{}{ // golang doesn't have builtin hash sets, so this is a workaround for that
+	"blob.core.windows.net":       nil,
```
vivlingaiah: shouldn't it be true instead of nil ?

D1v38om83r: Its inconsequential, I am only using the presence/absence of the key in the map/dictionary. I am not using the value.

PR #153, opened 2019-10-04
Fix for text file busy error for custom-script-extension binary

D1v38om83r:
Fix for text file busy custom-script-extension binary.
Use the shim file to detect open file handles to custom-script-extension binary, and kill the processes that own them before attempting to execute the binary.

### DIFF
--- a/misc/custom-script-shim
+++ b/misc/custom-script-shim
@@ -53,19 +53,22 @@ write_status() {
 check_binary_write_lock() {
     set +e # disable exit on non-zero return code
     local retry_attempts=0
-    while (( retry_attempts < 3 )); do
+    while (( retry_attempts < 10 )); do
         lsof_output="$(lsof ${bin})"
         if [ "$?" -eq 0 ]; then
             echo "${HANDLER_BIN} is open by the following processes: "
             echo "${lsof_output}"
             ((++retry_attempts))
-            echo "sleeping for 30 seconds before retry, attempt ${retry_attempts} of 3"
-            sleep 30
+            echo "sleeping for 3 seconds before retry, attempt ${retry_attempts} of 10"
+            sleep 3
         else
             set -e
             return 0 #Success path
         fi
     done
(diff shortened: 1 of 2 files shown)


misc/custom-script-shim
```diff
@@ -50,6 +50,19 @@ write_status() {
...
 
+check_binary_write_lock() {
+    set +e # disable exit on non-zero return code
+    lsof_output="$(lsof ${bin})"
+    if [ "$?" -eq 0 ]; then
+        echo "${HANDLER_BIN} is open by the following processes: "
+        echo "${lsof_output}"
+        echo "attempting to kill processes with open file handles to ${HANDLER_BIN}"
+        # suppress output and errors in case process with file-handle is already dead and kill gets called without a process id
+        lsof -t ${bin} | xargs kill -9 > /dev/null 2>&1
```
nkuchta: Given that this appears to be a race condition, I'm not sure we should do anything but log and add a bit of retry/wait logic here. There was speculation that waagent may be holding up the file - we don't want to kill that or any other critical process.

misc/custom-script-shim
```diff
@@ -50,6 +50,34 @@ write_status() {
...
+    set +e # disable exit on non-zero return code
+    local retry_attempts=0
+    while (( retry_attempts < 3 )); do
+        lsof_output="$(lsof ${bin})"
+        if [ "$?" -eq 0 ]; then
+            echo "${HANDLER_BIN} is open by the following processes: "
+            echo "${lsof_output}"
+            ((++retry_attempts))
+            echo "sleeping for 30 seconds before retry, attempt ${retry_attempts} of 3"
+            sleep 30
```
nkuchta: Increase the maximum retry attempts and decrease the interval between them (10 attempts, 5 seconds between?)

misc/custom-script-shim
nkuchta: For now, let's hold off on killing any processes and see if the retries fix the problem. With this change, we'll also gather data about which processes are the culprits.

PR #170, opened 2020-05-21
[Don't Merge] Making Azure blob validaiton more generic to support NCs

kamusta-msft:
Mirroring functionality in CSE Windows

### DIFF
--- a/pkg/download/blobwithmsitoken.go
+++ b/pkg/download/blobwithmsitoken.go
@@ -2,12 +2,13 @@ package download
 
 import (
 	"fmt"
-	"github.com/Azure/azure-extension-foundation/httputil"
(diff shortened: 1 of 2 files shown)


pkg/download/blobwithmsitoken.go
```diff
@@ -21,6 +21,8 @@ var azureBlobDomains = map[string]interface{}{ // golang doesn't have builtin ha
 	"blob.core.chinacloudapi.cn":  nil,
 	"blob.core.usgovcloudapi.net": nil,
 	"blob.core.couldapi.de":       nil,
+	".blob.core.":				   nil,
```
D1v38om83r, line 24: nit: make spacing uniform

PR #190, opened 2021-09-27
arm64 support and build update

koralski:


pkg/seqnum/seqnum.go
```diff
@@ -1,40 +0,0 @@
-// Package seqnum provides utilities to persistently store sequence number in a
```
D1v38om83r, line 1: Any good reason as to why the contents of this file were moved into main?

D1v38om83r (approved):

pkg/seqnum/seqnum.go
koralski: Reverted

PR #195, opened 2022-03-21
protected settings changes

norakoiralamsft:
-- clears out file in config folder (ex: 0.settings)


main/handlersettings_test.go
```diff
@@ -194,3 +200,48 @@ func Test_toJSONUmarshallForManagedIdentity(t *testing.T) {
...
+		content, err := ioutil.ReadFile(fileName)
+		assert.NoError(t, err)
+		assert.Equal(t, len(content), 0)
+	}
+
+	// cleanup
+	defer os.RemoveAll(testFolderPath)
+}
+
+func createTestFiles(folderPath, settingsExtensionName string) error {
```
jscalev, line 227: should add a test where we have one of the files open so we can't modify it (extension platform should have a similar test whose logic can be copied)

jscalev (changes requested):

main/handlersettings_test.go
```diff
@@ -194,3 +201,62 @@ func Test_toJSONUmarshallForManagedIdentity(t *testing.T) {
...
+	}
+
+	//verify that non settings file did not get cleared
+	fileName = filepath.Join(testFolderPath, "HandlerEnv.txt")
+	content, err := ioutil.ReadFile(fileName)
+	assert.NoError(t, err)
+	assert.Equal(t, len(content), 9)
+
+	// cleanup
+	defer os.RemoveAll(testFolderPath)
```
D1v38om83r, line 233: Can call lines with differ right after the Create TestFiles is called.

main/handlersettingscommon.go
```diff
@@ -140,3 +143,25 @@ func unmarshalProtectedSettings(configFolder string, hs handlerSettingsCommon, v
 	}
 	return nil
 }
+
+// cleanUpSettings replaces the protected settings for all settings files [ex: 0.settings, etc] to ensure no
+// protected settings are logged in VM
+func cleanUpSettings(ctx *log.Context, configFolder string) {
```
D1v38om83r, line 149: Would Azure-Extension-Platform be a better place to define this function in?

norakoiralamsft:
new PR: https://github.com/Azure/custom-script-extension-linux/pull/201

PR #199, opened 2022-09-01
run script on update

norakoiralamsft:
currently script is not being run when updating extension, 

--> currently sequence number is checked using status file, once that check is complete, added check if there's a new settings file (new sequence number) 
if there is a new setting file, then new seq number is written to mrseq file and the script should get executed

### DIFF
--- a/main/cmds.go
+++ b/main/cmds.go
@@ -130,11 +130,12 @@ func enablePre(ctx *log.Context, hEnv HandlerEnvironment, seqNum int) error {
 		return errors.Wrap(err, "failed to process sequence number")
 	} else if shouldExit {
 		//check if there's a new settings file. ex: ext updates and new script is passed in 
-	    newScript, err := checkNewSettings(ctx, hEnv, mostRecentSequence)
-		if !newScript || err != nil {
+		// and updates mostrecent seq number
+	    newScript := checkNewSettings(ctx, hEnv.HandlerEnvironment.ConfigFolder, mostRecentSequence)
+		if !newScript {
 			ctx.Log("event", "exit", "message", "the script configuration has already been processed, will not run again")
+			os.Exit(0)
 		} 
-		os.Exit(0)
 	}
 	return nil
 }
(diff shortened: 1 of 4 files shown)


main/cmds.go
```diff
@@ -123,14 +123,19 @@ func uninstall(ctx *log.Context, h HandlerEnvironment, seqNum int) (string, erro
...
 	// exit if this sequence number (a snapshot of the configuration) is already
 	// processed. if not, save this sequence number before proceeding.
 	if shouldExit, err := checkAndSaveSeqNum(ctx, seqNum, mostRecentSequence); err != nil {
 		return errors.Wrap(err, "failed to process sequence number")
 	} else if shouldExit {
-		ctx.Log("event", "exit", "message", "the script configuration has already been processed, will not run again")
-		os.Exit(0)
+		//check if there's a new settings file. ex: ext updates and new script is passed in 
+		// and updates mostrecent seq number
+	    newScript := checkNewSettings(ctx, hEnv.HandlerEnvironment.ConfigFolder, mostRecentSequence)
```
jscalev, line 134: I'm a bit confused on how this will resolve the issue. What happens is we seem to update the sequence number during install, and then during enable we think that sequence number has already run. Since we're saving the seqNo in checkAndSaveSeqNum above, I'm not sure how the checkNewSettings call will change this.

I suggest taking a look at the logs from the original event. It does seem that there's an issue when migrating settings. If mrseqno is deleted, then we write a new one with the seqNo sent to us from GA. However, we don't have a good idea whether that seqNo has ever run. If there is a GS in this seqNo, then we won't run it.

Please take a look whether this is what actually happened, or if I'm mistaken in the logic please let me know.

jscalev (changes requested):

main/cmds.go
norakoiralamsft: checkAndSaveSeqNum takes the sequence number from the order of the _status_ files. After that is checked,  checkNewSettings would check if there was a new _settings_ file (new settings for the script that was passed in), and set that as the latest seq number. The logic was that the settings file would have a higher seq number than the status file, so we would use that instead. Please let me know if I misunderstood how the files get ordered. 

In the meantime, I'll take a look at the logs from the original event to get a better understanding.

norakoiralamsft:
New changes are present here: https://github.com/Azure/custom-script-extension-linux/pull/200

PR #200, opened 2022-09-15
Adding logging + logic to migrating seq num

norakoiralamsft:
Adding logging: 
--> Wherever mrseq is being handled :) 

Logic added: 
--> sets seq num from environment variable in mrseq, if any errors with settings 
then uses status folder seq num isntead


main/cmds.go
```diff
@@ -92,22 +94,37 @@ func migrateToMostRecentSequence(ctx *log.Context, h HandlerEnvironment, seqNum
...
-		ctx.Log("event", "migrate to mrseq", "error", err)
-		return
+	if seqNum != 0 {
+		computedSeqNum := seqNum
+		_, err := fout.WriteString(fmt.Sprintf("%v", computedSeqNum))
+		if err == nil {
+			ctx.Log("event", "migrated to mrseq", "message", fmt.Sprintf("migrated mrseq to %v from config seqnum", computedSeqNum))
+			return 
+		} 
+		//if error writing seqNum, then try getting seq num from status
```
jscalev, line 111: so, if the above generates an error, that means we couldn't write the new mrseq, right? In that case, I don't think it will help to attempt to recompute it, because that's not where the problem was. Clearly something was holding a lock to prevent us from writing, so we should probably write out the error in that case and presumably fail the install since we can no longer guarantee the seqNo.

main/cmds.go
```diff
@@ -92,22 +94,37 @@ func migrateToMostRecentSequence(ctx *log.Context, h HandlerEnvironment, seqNum
...
 
-	fout, err := os.Create(mostRecentSequence)
-	if err != nil {
-		ctx.Log("event", "migrate to mrseq", "error", err)
-		return
+	if seqNum != 0 {
+		computedSeqNum := seqNum
+		_, err := fout.WriteString(fmt.Sprintf("%v", computedSeqNum))
+		if err == nil {
+			ctx.Log("event", "migrated to mrseq", "message", fmt.Sprintf("migrated mrseq to %v from config seqnum", computedSeqNum))
```
jscalev, line 108: We're not really migrating, right? Mrseq was deleted, so we're doing our best to recreate it.

jscalev (changes requested):

main/cmds.go
```diff
@@ -92,22 +94,37 @@ func migrateToMostRecentSequence(ctx *log.Context, h HandlerEnvironment, seqNum
...
-		// If there was an error, the sequence number is zero.
 		ctx.Log("event", "migrate to mrseq", "error", err)
-		return
 	}
 
-	fout, err := os.Create(mostRecentSequence)
-	if err != nil {
-		ctx.Log("event", "migrate to mrseq", "error", err)
-		return
+	if seqNum != 0 {
```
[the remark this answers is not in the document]
norakoiralamsft: Yes, 0 is a vaild seqNum. When seqNum is 0, it would be the first installation so we should not write to mrseq during install. We later check the seqNum against mrseq during pre-enable and ext will assume that seqNum has already been run. At pre-enable, mrseq will be updated accordingly.

main/cmds.go
norakoiralamsft: That makes sense. In that case, I think that seqNo should always have some value to it based on the logic in main.go, so would checking status folders still be necessary?

main/cmds.go
norakoiralamsft: Yup... was just keeping current log format. Maybe, "recreating mrseq" would be better?

main/cmds.go
```diff
@@ -79,6 +79,8 @@ func install(ctx *log.Context, h HandlerEnvironment, seqNum int) (string, error)
 }
 
 func migrateToMostRecentSequence(ctx *log.Context, h HandlerEnvironment, seqNum int) {
+	// First the config seq number is used for mrseq. If there is an error saving that, then...
```
vivlingaiah, line 82: "config seq number" looks like a description. What's the env variable name ? Also, please mention what component saves it.

main/cmds.go
```diff
@@ -92,22 +94,37 @@ func migrateToMostRecentSequence(ctx *log.Context, h HandlerEnvironment, seqNum
 	// do not have invent another method.  The CustomScript extension should have been using this
 	// from the beginning, but it was not.
 	//
-	computedSeqNum, err := FindSeqNumStatus(h.HandlerEnvironment.StatusFolder)
+	ctx.Log("event", "migrate to mrseq", "message", fmt.Sprintf("cannot find mrseq at %v", mostRecentSequence))
+	fout, err := os.Create(mostRecentSequence)
+	defer fout.Close()
 	if err != nil {
```
D1v38om83r, line 100: If there is a file creation error at this point, it is very likely all subsequent operations with fout would fail, or even throw a null reference panic. We used to exit early here. Why the change to continue after os.Create(mostRecentSequence) failed?

main/cmds.go
D1v38om83r: func migrateToMostRecentSequence should also return an error if it fails to write to mrseq file and fail the installation.

main/cmds.go
```diff
@@ -79,6 +79,8 @@ func install(ctx *log.Context, h HandlerEnvironment, seqNum int) (string, error)
 }
 
 func migrateToMostRecentSequence(ctx *log.Context, h HandlerEnvironment, seqNum int) {
```
D1v38om83r, line 81: Please add a unit test for func migrateToMostRecentSequence under cmds_test.go

D1v38om83r (changes requested):

D1v38om83r (approved):

main/cmds.go
```diff
@@ -72,44 +72,40 @@ func install(ctx *log.Context, h HandlerEnvironment, seqNum int) (string, error)
...
-	// do not have invent another method.  The CustomScript extension should have been using this
-	// from the beginning, but it was not.
-	//
-	computedSeqNum, err := FindSeqNumStatus(h.HandlerEnvironment.StatusFolder)
+func migrateToMostRecentSequence(ctx *log.Context, seqNum int) error {
+	// The configSequenceNumber (seqNum) env variable is used for mrseq if mrseq not found
+	// based on logic in main.go the seqNum should always be populated
+	ctx.Log("event", "recreate mrseq", "message", fmt.Sprintf("cannot find mrseq at %v", mostRecentSequence))
+	fout, err := os.Create(mostRecentSequence)
+	defer fout.Close()
```
jscalev, line 91: We shouldn't defer fout.Close() if there's an error, so this should probably be moved below the error check

main/cmds.go
```diff
@@ -72,44 +72,40 @@ func install(ctx *log.Context, h HandlerEnvironment, seqNum int) (string, error)
...
 	if err != nil {
-		// If there was an error, the sequence number is zero.
-		ctx.Log("event", "migrate to mrseq", "error", err)
-		return
+		ctx.Log("event", "recreate mrseq", "error", err)
```
jscalev, line 93: "Error creating sequence file"

main/cmds.go
```diff
@@ -72,44 +72,40 @@ func install(ctx *log.Context, h HandlerEnvironment, seqNum int) (string, error)
 	// If it is (2) attempt to migrate to mrseq.  Find the latest settings
 	// file, and set the sequence to it.
 	if _, err := os.Stat(mostRecentSequence); os.IsNotExist(err) {
-		migrateToMostRecentSequence(ctx, h, seqNum)
+		err := migrateToMostRecentSequence(ctx, seqNum)
```
jscalev, line 75: Previously, install would never fail for this extension. Given how often this extension is used, I recommend keeping the current behavior since this change is only for logging.

jscalev (approved):

PR #201, opened 2022-09-15
Clear protected settings

norakoiralamsft:
-- Clears protected settings files in config folder


D1v38om83r (approved):
Please add vendor directory to .gitignore, and not check it in with this codebase.

PR #203, opened 2023-01-11
Track Request ID, Response ID, and Response Code + Add response-code-specific error messages

deeptivaidymsft:
For [Error Improvements in VM Extensions](https://microsoft.sharepoint.com/:w:/t/ComputeVM/ERQYqh3QTktFrdtAgBMZvAgBLKT8DqKH3K0rKznDVCr11A?e=bJE1Zm) , when downloading scripts we will log the request ID, response ID, and response code (if provided) and give specific error messages for status codes 401, 404, 400, 500.

### DIFF
--- a/pkg/download/downloader.go
+++ b/pkg/download/downloader.go
@@ -49,7 +49,6 @@ func Download(ctx *log.Context, d Downloader) (int, io.ReadCloser, error) {
 	if err != nil {
 		return -1, nil, errors.Wrapf(err, "failed to create http request")
 	}
-
 	requestID := req.Header.Get(xMsClientRequestIdHeaderName)
 	if len(requestID) > 0 {
 		ctx.Log("info", fmt.Sprintf("starting download with client request ID %s", requestID))
@@ -76,29 +75,34 @@ func Download(ctx *log.Context, d Downloader) (int, io.ReadCloser, error) {
 		}
 		break
 	default:
+		uri := req.URL.Host //"string"
 		switch resp.StatusCode {
 		case http.StatusUnauthorized:
-			errString = fmt.Sprintf("CustomScript failed to download the blob [REDACTED] because access was denied. Please fix the blob permissions and try again, the response code returned was: %q",
-				resp.Status,
-			)
+			errString = fmt.Sprintf("CustomScript failed to download the file from %s because access was denied. Please fix the blob permissions and try again, the response code and message returned were: %q",
+				uri,
+				resp.Status)
 		case http.StatusNotFound:
-			errString = fmt.Sprintf("CustomScript failed to download the blob [REDACTED] because it does not exist. Please create the blob and try again, the response code returned was %q",
+			errString = fmt.Sprintf("CustomScript failed to download the file from %s because it does not exist. Please create the blob and try again, the response code and message returned were: %q",
+				uri,
 				resp.Status)
 
 		case http.StatusBadRequest:
-			errString = fmt.Sprintf("CustomScript failed to download the blob [REDACTED] because parts of the request were incorrectly formatted, missing, and/or invalid. The response code returned was: %q",
+			errString = fmt.Sprintf("CustomScript failed to download the file from %s because parts of the request were incorrectly formatted, missing, and/or invalid. The response code and message returned were: %q",
+				uri,
 				resp.Status)
 
 		case http.StatusInternalServerError:
-			errString = fmt.Sprintf("CustomScript failed to download the blob [REDACTED] due to an issue with storage. The response code returned was: %q",
+			errString = fmt.Sprintf("CustomScript failed to download the file from %s due to an issue with storage. The response code and message returned were: %q",
+				uri,
 				resp.Status)
 		default:
-			errString = fmt.Sprintf("CustomScript failed to download the blob [REDACTED] because the server returned a response of %q. Please verify the machine has network connectivity.",
+			errString = fmt.Sprintf("CustomScript failed to download the file from %s because the server returned a response code and message of %q Please verify the machine has network connectivity.",
+				uri,
 				resp.Status)
 		}
 	}
 	if len(requestId) > 0 {
-		errString += fmt.Sprintf(" (Service request ID: %s", requestId)
+		errString += fmt.Sprintf(" (Service request ID: %s)", requestId)
 	}
 	return resp.StatusCode, nil, fmt.Errorf(errString)
 }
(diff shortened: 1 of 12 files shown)


pkg/download/downloader.go
```diff
@@ -58,15 +64,41 @@ func Download(d Downloader) (int, io.ReadCloser, error) {
...
+			errString = MsiDownload404ErrorString
 		case http.StatusForbidden:
-			return resp.StatusCode, nil, errors.Wrapf(err, MsiDownload403ErrorString)
+			errString = MsiDownload403ErrorString
+		}
+		break
+	default:
+		switch resp.StatusCode {
+		case http.StatusUnauthorized:
+			errString = fmt.Sprintf("CustomScript failed to download the blob [REDACTED] because access was denied. Please fix the blob permissions and try again, the response code returned was: %q",
```
vivlingaiah, line 81: You could mention what blob permissions are needed to make it easier for user.

pkg/download/downloader.go
```diff
@@ -58,15 +64,41 @@ func Download(d Downloader) (int, io.ReadCloser, error) {
...
+				resp.Status,
+			)
+		case http.StatusNotFound:
+			errString = fmt.Sprintf("CustomScript failed to download the blob [REDACTED] because it does not exist. Please create the blob and try again, the response code returned was %q",
```
vivlingaiah, line 85: [REDACTED] - Is there any issue mentioning base blob URL ? It'd help user to know which blob they have an issue with.

pkg/download/downloader.go
```diff
@@ -58,15 +64,41 @@ func Download(d Downloader) (int, io.ReadCloser, error) {
...
+				resp.Status)
+
+		case http.StatusBadRequest:
+			errString = fmt.Sprintf("CustomScript failed to download the blob [REDACTED] because parts of the request were incorrectly formatted, missing, and/or invalid. The response code returned was: %q",
+				resp.Status)
+
+		case http.StatusInternalServerError:
+			errString = fmt.Sprintf("CustomScript failed to download the blob [REDACTED] due to an issue with storage. The response code returned was: %q",
```
vivlingaiah, line 93: Is it possible to include the error message returned from the downstream service like storage if it is available in response ? That'd help us identify issue faster.

pkg/download/downloader.go
```diff
@@ -58,15 +64,41 @@ func Download(d Downloader) (int, io.ReadCloser, error) {
...
+				resp.Status)
+		default:
+			errString = fmt.Sprintf("CustomScript failed to download the blob [REDACTED] because the server returned a response of %q. Please verify the machine has network connectivity.",
```
vivlingaiah, line 96: consider including an error message if available in response

pkg/download/downloader_test.go
```diff
@@ -46,17 +52,29 @@ func TestDownload_badStatusCodeFails(t *testing.T) {
 		http.StatusBadRequest,
 		http.StatusUnauthorized,
 	} {
-		_, _, err := download.Download(download.NewURLDownload(fmt.Sprintf("%s/status/%d", srv.URL, code)))
+		_, _, err := download.Download(testctx, download.NewURLDownload(fmt.Sprintf("%s/status/%d", srv.URL, code)))
```
vivlingaiah, line 55: This looks a bit confusing. What does this test try to achieve ? It has all response codes listed. What is the scenario and what is expected status code ?

pkg/download/downloader_test.go
```diff
@@ -46,17 +52,29 @@ func TestDownload_badStatusCodeFails(t *testing.T) {
...
 		require.NotNil(t, err, "not failed for code:%d", code)
-		require.Contains(t, err.Error(), "unexpected status code", "wrong message for code %d", code)
+		switch code {
+		case http.StatusNotFound:
+			require.Contains(t, err.Error(), "because it does not exist")
```
vivlingaiah, line 59: could you also assert for status code returned using require.AreEqual(<expectedHttpStatusCode>, <actualHttpStatusCode>)

pkg/download/downloader.go
deeptivaidymsft: Yes, the response code and message are both in resp.Status

pkg/download/downloader_test.go
deeptivaidymsft: The purpose of this test is to ensure that the error message returned is formatted correctly based on the specific error code. I didn't think it was necessary to create several small tests for this, but I can change the name of the test so its purpose is more clear

jscalev (approved):

pkg/download/downloader.go
```diff
@@ -58,15 +63,46 @@ func Download(d Downloader) (int, io.ReadCloser, error) {
...
+				uri,
+				resp.Status)
+
+		case http.StatusInternalServerError:
+			errString = fmt.Sprintf("CustomScript failed to download the file from %s due to an issue with storage. The response code and message returned were: %q",
+				uri,
+				resp.Status)
+		default:
+			errString = fmt.Sprintf("CustomScript failed to download the file from %s because the server returned a response code and message of %q Please verify the machine has network connectivity.",
+				uri,
```
malhotrarohit, line 100: does uri also contain path? a typical uri looks like https://hostname/path?query
AFAIK scripts can be inside blob containers and container name is usually in the path.
I believe CSE creates a directory with the same name as the blob container before downloading the blob, that is in case blob is inside a container.

pkg/download/downloader.go
```diff
@@ -58,15 +63,46 @@ func Download(d Downloader) (int, io.ReadCloser, error) {
...
+				resp.Status)
+		case http.StatusNotFound:
+			errString = fmt.Sprintf("CustomScript failed to download the file from %s because it does not exist. Please create the blob and try again, the response code and message returned were: %q",
+				uri,
+				resp.Status)
+
+		case http.StatusBadRequest:
+			errString = fmt.Sprintf("CustomScript failed to download the file from %s because parts of the request were incorrectly formatted, missing, and/or invalid. The response code and message returned were: %q",
+				uri,
+				resp.Status)
```
malhotrarohit, line 92: does the status also contain status code or is it just the response body?

pkg/download/downloader.go
deeptivaidymsft: uri only shows the host or host:port part of the request URL. I'll rename the variable since it's a bit misleading, but the path and everything after should be cut off

PR #211, opened 2023-04-13
Deleting scripts and updating settings file logic

norakoiralamsft:
Deleting scripts except for the last one, and updating settings files except for the last one


main/cmds.go
```diff
@@ -173,9 +188,21 @@ func enable(ctx *log.Context, h HandlerEnvironment, seqNum int) (string, error)
 		ctx.Log("event", "enable failed")
 	}
 
-	ctx.Log("event", "clearing setting files")
-	el := logging.New(nil)
-	settings.CleanUpSettings(el, h.HandlerEnvironment.ConfigFolder)
+	// clearing settings and scripts
```
malhotrarohit, line 191: you could potentially isolate and put these lines (that are being repeated) in a dedicated method and call it from enablePre and enable methods, something like deleteSettingsAndScriptsForAllSequenceNumbersExceptThisOne(seqNum int)

main/cmds.go
```diff
@@ -124,8 +124,22 @@ func enablePre(ctx *log.Context, hEnv HandlerEnvironment, seqNum int) error {
 		return errors.Wrap(err, "failed to process sequence number")
 	} else if shouldExit {
```
malhotrarohit, line 125: I suppose you should add tests to make sure that the new behavior you're adding is working as expected

main/cmds.go
```diff
@@ -124,8 +124,22 @@ func enablePre(ctx *log.Context, hEnv HandlerEnvironment, seqNum int) error {
...
-		settings.CleanUpSettings(el, hEnv.HandlerEnvironment.ConfigFolder)
+		// clearing settings and scripts
+		ctx.Log("event", "clearing settings and script files")
+		downloadsParent := filepath.Join(dataDir, downloadDir)
+		seqNumString := strconv.Itoa(seqNum)
+		err = utils.TryDeleteDirectoriesExcept(downloadsParent, seqNumString)
+		if err != nil {
+			ctx.Log("event", "could not clear scripts")
+		}
+		mostRecentRuntimeSetting := fmt.Sprintf("%d.settings", uint(seqNum))
```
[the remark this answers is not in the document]
D1v38om83r: Should be inconsequential unless the sequence number gets to a value larger than 2^31

D1v38om83r (commented):
Please add unit tests and separate the method that clears out the directories so that unit testing is easier.

main/cmds.go
norakoiralamsft: Tests are in this [repo](https://github.com/Azure/azure-extension-platform/blob/main/pkg/utils/utils_linux_test.go) 

Should I add similar ones here?

main/cmds.go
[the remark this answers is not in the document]
norakoiralamsft: Don't think golang supports %u, unless I'm misunderstanding: [golang docs](https://pkg.go.dev/fmt )

main/cmds.go
norakoiralamsft: Added.

main/cmds.go
```diff
@@ -173,9 +188,21 @@ func enable(ctx *log.Context, h HandlerEnvironment, seqNum int) (string, error)
...
-	el := logging.New(nil)
-	settings.CleanUpSettings(el, h.HandlerEnvironment.ConfigFolder)
+	// clearing settings and scripts
+	ctx.Log("event", "clearing settings and script files")
+	seqNumString := strconv.Itoa(seqNum)
+	err = utils.TryDeleteDirectoriesExcept(downloadsParent, seqNumString)
+	if err != nil {
+		ctx.Log("event", "could not clear scripts")
+	}
+	mostRecentRuntimeSetting := fmt.Sprintf("%d.settings", uint(seqNum))
```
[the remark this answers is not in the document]
norakoiralamsft: Same as above.

main/cmds.go
```diff
@@ -339,3 +337,22 @@ func decodeScript(script string) (string, string, error) {
 	w.Flush()
 	return buf.String(), fmt.Sprintf("%d;%d;gzip=1", len(script), n), nil
 }
+
+func clearSettingsAndScriptExceptMostRecent(seqNum) {
+	downloadsParent := filepath.Join(dataDir, downloadDir)
+	seqNumString := strconv.Itoa(seqNum)
+
+	ctx.Log("event", "clearing settings and script files except most recent seq num")
+	err = utils.TryDeleteDirectoriesExcept(downloadsParent, seqNumString)
```
malhotrarohit, line 346: can TryDeleteDirectoriesExcept and TryClearRegexMatchingFilesExcept return error even if some directories/files are deleted and some aren't? in that case it would help to log which files could not be deleted. Logging a single aggregate message such as "could not clear scripts" is less useful than logging the names of files that could not be deleted. Note that this can be done inside of the Try* methods and you don't necessarily have to return a list of file/directory names to their caller. The caller can simply log "some scripts could not be deleted" or "some settings files could not be deleted"

malhotrarohit (approved):

main/cmds.go
norakoiralamsft: I'll address this in a separate PR in the common repo!

PR #212, opened 2023-04-21
Include the entire read response body and file save operation inside retry method

malhotrarohit:
What operation are we performing on behalf of the customer?
- Downloading the script files and saving them to locla disk

What part of this operation are we retrying on?
- Just the HTTP download and that too only if status is not 200 

What should we be retrying ideally?
- The entire operation  = HTTP request + streaming response + writing streamed bytes to local disk

This PR aims to make all parts of the operation retryable.

### DIFF
--- a/pkg/download/retry.go
+++ b/pkg/download/retry.go
@@ -47,7 +47,7 @@ func WithRetries(ctx *log.Context, f *File, downloaders []Downloader, sf SleepFu
 				// we have a response body, copy it to the file
 				nBytes, err := io.CopyBuffer(f, out, make([]byte, writeBufSize))
 				if err == nil {
-					// we are done, close the response body
+					// we are done, close the response body, log time taken to download the file
 					// and return the number of bytes written
 					out.Close()
 					end = time.Since(start)
@@ -64,6 +64,9 @@ func WithRetries(ctx *log.Context, f *File, downloaders []Downloader, sf SleepFu
 				}
 			}
 
+			// we are here because either server returned a non-200 status code
+			// or we failed to download the response body and write it to file
+			// log the error, time elapsed, bytes downloaded, and close the response body
 			end = time.Since(start)
 			ctx.Log("error", fmt.Sprintf("file download failed with error '%s' : downloaded %d bytes in %d milliseconds", err, nBytes, end.Milliseconds())
 
--- a/pkg/download/retry.go
+++ b/pkg/download/retry.go
@@ -5,6 +5,7 @@ import (
 	"io"
 	"math"
 	"net/http"
+	"os"
 	"time"
 
 	"github.com/go-kit/kit/log"
@@ -24,6 +25,7 @@ const (
 	expRetryN = 7 // how many times we retry the Download
 	expRetryK = time.Second * 3
 	expRetryM = 2
+	writeBufSize = 1024 * 8
 )
 
 // WithRetries retrieves a response body using the specified downloader. Any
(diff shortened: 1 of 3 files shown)


pkg/download/retry.go
```diff
@@ -24,21 +25,36 @@ const (
...
+	writeBufSize = 1024 * 8
 )
 
 // WithRetries retrieves a response body using the specified downloader. Any
 // error returned from d will be retried (and retrieved response bodies will be
 // closed on failures). If the retries do not succeed, the last error is returned.
 //
 // It sleeps in exponentially increasing durations between retries.
-func WithRetries(ctx *log.Context, downloaders []Downloader, sf SleepFunc) (io.ReadCloser, error) {
+func WithRetries(ctx *log.Context, f *File, downloaders []Downloader, sf SleepFunc) (int64, error) {
```
jscalev, line 36: Please double check whether there are other callers for this package. I believe RunCommand V2 may use it, in which case it will break when the pkg is next imported.

pkg/download/retry.go
```diff
@@ -24,21 +25,36 @@ const (
...
+				if err == nil {
+					// we are done, close the response body
+					// and return the number of bytes written
+					out.Close()
+					return nBytes, nil
+				}
+				else {	
+					// we failed to download the response body and write it to file
+					// because either connection was closed prematurely or file write operation failed
+					// mark status as -1 so that we retry
```
jscalev, line 55: should probably ctx.Log here

pkg/download/retry.go
```diff
@@ -24,21 +25,36 @@ const (
...
-func WithRetries(ctx *log.Context, downloaders []Downloader, sf SleepFunc) (io.ReadCloser, error) {
+func WithRetries(ctx *log.Context, f *File, downloaders []Downloader, sf SleepFunc) (int64, error) {
 	var lastErr error
 	for _, d := range downloaders {
 		for n := 0; n < expRetryN; n++ {
 			ctx := ctx.With("retry", n)
 			status, out, err := Download(ctx, d)
 			if err == nil {
-				return out, nil
+				// server returned status code 200 OK
```
jscalev, line 43: We do appear to have unit tests under retry_test, so we should add one/some for this case.

jscalev (changes requested):

pkg/download/retry.go
```diff
@@ -24,21 +25,36 @@ const (
...
+				// we have a response body, copy it to the file
+				nBytes, err := io.CopyBuffer(f, out, make([]byte, writeBufSize))
```
D1v38om83r, line 45: tests to ensure that the file is properly overwritten and not incorrectly appended due to f *File being a stream.

pkg/download/retry.go
D1v38om83r: Imports are versioned, importers can get away with not having to change their code as long as don't update their reference to this repo in their go.mod file via some command like "go get -u github.com/Azure/custom-script-extension-linux"

pkg/download/retry.go
D1v38om83r, line 45: also can we log the content size of out to GuestAgentExtension and local log? Time the download as well.

pkg/download/retry.go
malhotrarohit: Looks like other extensions have their own implementations of Save, Download, and WithRetries methods  -

https://github.com/Azure/run-command-extension-linux/blob/master/pkg/download/save.go
https://github.com/Azure/run-command-extension-linux/blob/master/pkg/download/retry.go

https://github.com/Azure/run-command-handler-linux/blob/main/pkg/download/save.go
https://github.com/Azure/run-command-handler-linux/blob/main/pkg/download/retry.go

pkg/download/retry.go
malhotrarohit: We're already calling ctx.log in line # 63, so I don't think we need to call it here.

pkg/download/retry.go
malhotrarohit: I've added a call to f.truncate in else condition (err != nil) so that we will truncate the file after every failure in io.CopyBuffer.

jscalev (approved):

PR #214, opened 2023-05-02
Reverting PR 200 -- seqNum fix 

norakoiralamsft:
PR 200 doesn't address the issue directly and introduces a new bug, will revert to unblock current release.

### DIFF
--- a/main/cmds.go
+++ b/main/cmds.go
@@ -72,7 +72,7 @@ func install(ctx *log.Context, h HandlerEnvironment, seqNum int) (string, error)
 	// If it is (2) attempt to migrate to mrseq.  Find the latest settings
 	// file, and set the sequence to it.
 	if _, err := os.Stat(mostRecentSequence); os.IsNotExist(err) {
-		migrateToMostRecentSequence(ctx, seqNum)
+		migrateToMostRecentSequence(ctx, h, seqNum)
 	}
 
 	ctx.Log("event", "created data dir", "path", dataDir)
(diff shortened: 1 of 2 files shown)


main/cmds.go
```diff
@@ -72,36 +72,44 @@ func install(ctx *log.Context, h HandlerEnvironment, seqNum int) (string, error)
...
 }
 
-func migrateToMostRecentSequence(ctx *log.Context, seqNum int) {
-	// The configSequenceNumber (seqNum) env variable is used for mrseq if mrseq not found
-	// based on logic in main.go the seqNum should always be populated
-	ctx.Log("event", "recreate mrseq", "message", fmt.Sprintf("cannot find %v", mostRecentSequence))
-	fout, err := os.Create(mostRecentSequence)
+func migrateToMostRecentSequence(ctx *log.Context, h HandlerEnvironment, seqNum int) {
+	// The status folder is used instead of the settings because the settings file is written
+	// by the agent before install is called.  As a result, the extension cannot determine if this
```
malhotrarohit, line 85: why is settings file being written by the guest agent before install is called, a problem? is update command called before install? so you have to rely on #.status to know the most recent seq num processed by the previous version of the extn.?

D1v38om83r (approved):

main/cmds.go
norakoiralamsft: Yes exactly, the goal is to get the last run seqnum in the case where the mrseq file is not present. The bug I was originally trying to fix was that the status gets written during update so when we get the seqnum it's the most recent one and not the previous one. One way I was thinking of fixing this is saving the previous seqnum during update scenarios, but open to any ideas you have!

PR #216, opened 2023-05-23
Removing migrate to mostRecentSeqNum logic

norakoiralamsft:
By removing this logic: 
- Upgrades that do not have mrseq file present will cause the currentSeqNum to run. 
- Upgrades that do have mrseq file present will honor the number in the mrseq file and only execute if mrseqNum less than the currentSeqNum.
- If it is the first ever install, the script will also run since there is no mrseq file to compare currentSeqNum to.

### DIFF
--- a/main/cmds.go
+++ b/main/cmds.go
@@ -77,38 +77,6 @@ func install(ctx *log.Context, h HandlerEnvironment, seqNum int) (string, error)
 	return "", nil
 }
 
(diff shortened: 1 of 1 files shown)


jscalev (approved):

D1v38om83r (approved):
Is there a reason to leave the old code is as comments?

PR #259, opened 2026-07-10
CSE ignores invalid policy file

alsanmsft:
Overview:
Part of the constrained extensions code involves reading a policy file (the file that contains the policy rules for the extension). Previously, if the file was corrupt (invalid JSON, empty/cleared file), then the extension would log the error and exit.

However, all the constrained extensions feature work is new, and we should err on the side of caution. Thus, for now if the policy file is invalid, we log the error but continue execution with no policy applied. This change matches the code changes made [here](https://dev.azure.com/msazure/One/_git/Compute-ART-Extensions?path=/src/CustomScriptHandler/src/Program.cs&version=GBmaster&line=307&lineEnd=314&lineStartColumn=14&lineEndColumn=14&lineStyle=plain&_a=contents) in CSE Windows.

PR includes:
* error log for when policy file invalid + some refactoring logic
* UT to ensure extension still succeeds + script is executed


main/cmds_test.go
```diff
@@ -152,6 +152,55 @@ func Test_LoadExtensionPolicySettings_InvalidFormat_RequireSigningWithoutRootCA(
...
+					}
+				}
+			}
+		]
+	}`
+	settingsFilePath := filepath.Join(configFolder, fmt.Sprintf("%d.settings", seqNum))
+	require.Nil(t, writeToFile(settingsFilePath, settingsContent))
+
+	// Write an invalid extension policy settings file that fails to load.
+	invalidPolicyContent := `{
```
D1v38om83r, line 191: will this cause extensionPolicyManagerPtr.LoadExtensionPolicySettings() or extensionPolicyManagerPtr.GetSettings() to return error?

should we include examples that exercises both new code paths one after the other?

D1v38om83r (approved):

PR #261, opened 2026-08-20
cmd to execute and script are limited by policy

alsanmsft:
**Overview/Problem:**

As part of the constrained extensions work, CSE has the option to limit what scripts are run. This was originally implemented as limiting the scripts downloaded, but that design can be easily circumvented by placing the full script into the "commandToExecute" field. 

This PR also includes some name changes for clarity

**Solution:**

Thus, this PR implements a commandToExecute allowlist and a Script allowlist (inline script). After the commandToExecute is parsed, the extension checks to see if policy settings will be applied (if they are present + have been set correctly). If so, then it checks to see if a commandToExecute allowlist is present and non-empty; if so, it compares the provided commandToExecute to the allowlist. If the list is non-empty and the command is not present, the extension terminates and exits with an error. If commandtoExecute is not available, and the 'script' parameter has been used instead, it also validates that value against the script allowlist in the policy.

CSE Windows corresponding PR: https://dev.azure.com/msazure/One/_git/Compute-ART-Extensions/pullrequest/16713202


main/cmds.go
```diff
@@ -301,6 +316,23 @@ func runCmd(ctx log.Logger, dir string, cfg handlerSettings) (ewc *vmextension.E
...
 
+// While this function usually validates the 'commandToExecute' property value against the policy,
+// if the 'script' property is used instead, it will validate that against the policy.
+// Only call this function if settings is not nil.
+// Note: the comparison is case sensitive! Trailing and leading whitespace is trimmed, but case is not ignored.
+func validateCommandToExecuteAgainstPolicy(commandToExecute string, list []string, scenario string) *vmextension.ErrorWithClarification {
+	if len(list) > 0 {
+		err := extensionpolicysettings.ValidateValueInAllowlist(commandToExecute, list)
+		if err != nil {
+			if strings.HasPrefix(scenario, "protected-script") {
```
jscalev: Since we already checked this before calling, we can just pass it as a bool

main/cmds.go
```diff
@@ -286,6 +290,17 @@ func runCmd(ctx log.Logger, dir string, cfg handlerSettings) (ewc *vmextension.E
 		scenario = fmt.Sprintf("protected-script;%s", scenarioInfo)
 	}
 
+	if settings != nil {
+		if strings.HasPrefix(scenario, "protected-script") {
```
jscalev: How/where is this prefix being set? Worth having a comment to explain.

alsanmsft: was getting it from line 290 is runCmd(), but I'm now longer using the prefix

D1v38om83r (approved):
LGMT

jscalev (approved):

PR #263, opened 2026-08-26
Custom Script shim

FeyiAdeaga:
Updated Custom script to use Fuser instead of Lsof.


misc/custom-script-shim
```diff
@@ -122,17 +81,58 @@ exec &> >(tee -ia "$LOG_DIR/$LOG_FILE")
...
     # after the 15-minute extension enabling timeout.
+    # & will execute the binary on the background and will not block current shell execution
     write_status
-    check_binary_write_lock
-    set -x
-    nohup "$bin" $@ &
+    commandToExecute="nohup '$bin' '$cmd' &"
+fi
+
+### Retry logic to check if RunCommand binary is open for write and execute the command
```
norakoiralamsft: nit: some comments still refer to run command binary

misc/custom-script-shim
```diff
@@ -122,17 +81,58 @@ exec &> >(tee -ia "$LOG_DIR/$LOG_FILE")
...
+            echo "Unexpected Fuser error with return code $fuser_return_code while checking if $bin is open for write.using fuser command. Proceeding with best-effort execution..."
+            break
+        fi
+    done
 else
-    # execute the handler process as a child process
-    check_binary_write_lock
-    set -x
-    "$bin" $@
+    echo "INFO: Not an error. fuser command not found. Install 'psmic' package to use fuser for checking if the Run Command binary is open for write to detect potential conflicts like 'text file busy' errors while executing Run Command binary. Proceeding with best-effort execution..."
```
norakoiralamsft: nit: change run command binary to cse

FeyiAdeaga: Resolved

misc/custom-script-shim
FeyiAdeaga: Resolved

misc/custom-script-shim
```diff
@@ -122,17 +81,58 @@ exec &> >(tee -ia "$LOG_DIR/$LOG_FILE")
...
+retry_attempts=0
+
+set -x
+
+if command -v fuser&>/dev/null; then
+    while (( retry_attempts < 10 )); do
+        # Use fuser to detect if the Custom script binary is open for write to help detect potential conflicts like 'text file busy' errors that prevents execution of the Custom script binary.
+        fuser_output=$(fuser -vw "$bin" 2>&1)
+        fuser_status=$?
+        fuser_output=$(fuser -v "$bin" 2>&1)
```
norakoiralamsft, line 106: Why is fuser_output being captured twice?

misc/custom-script-shim
```diff
@@ -122,17 +81,58 @@ exec &> >(tee -ia "$LOG_DIR/$LOG_FILE")
...
+    # & will execute the binary on the background and will not block current shell execution
     write_status
-    check_binary_write_lock
-    set -x
-    nohup "$bin" $@ &
+    commandToExecute="nohup '$bin' '$cmd' &"
+fi
+
+### Retry logic to check if Custom script binary is open for write and execute the command
+set -e
```
norakoiralamsft, line 97: if there's set -e and fuser returns an error code (for example "1"), it will return immediately. set +e so we can catch the error and handle in 119

misc/custom-script-shim
FeyiAdeaga: Resolved. Left over on previous code

misc/custom-script-shim
FeyiAdeaga: Resolved

norakoiralamsft: It's still set -e, should be set +e

norakoiralamsft (approved):

